PR #3, opened 2020-02-29
Draft library API

mijicd:
### Summary

- Defines library building blocks.
- Exposes an API for some of the commonly used command groups.
- Aligns CircleCI config with the rest of ZIO organization.

### Limitations

- Commands aren't interpreted (no transport is implemented). This will be done in a separate PR.
- Some command groups aren't supported and follow-up issues will be opened to tackle them.


mijicd:
@jdegoes tagging you just to get the preliminary feedback, as discussed before. There are still some rough edges, and I'd like to have sorted sets API before this can be even considered as ready to merge.

src/main/scala/zio/redis/Command.scala
```diff
@@ -0,0 +1,83 @@
...
+import scala.util.matching.Regex
+
+final case class Command[-In, +Out] private[redis] (
+  name: String,
+  input: Command.Input[In],
+  output: Command.Output[Out]
+) {
+
+  // main command interpreter
+  def run(in: In): Out = ???
```
paulpdaniels: How will the dependencies get passed in? In order to run a command, we'll likely need an actual `Redis` backed command interpreter, but all of the commands output `IO[Error, A]` instead of `ZIO[Redis, Error, A]`

src/main/scala/zio/redis/Command.scala
```diff
@@ -0,0 +1,83 @@
...
+
+import java.time.Instant
+
+import zio.{ Chunk, IO }
+import zio.stream.Stream
+import zio.duration.Duration
+
+import scala.util.matching.Regex
+
+final case class Command[-In, +Out] private[redis] (
```
joeyfeldberg: How will pipelined commands be handled with this API? pipelined commands can either return nothing if they are queued in memory or they can be sent and will return QUEUED from Redis as a simple string, most Scala libraries I've seen do the ugly thing of returning null (as instances of the return type).

src/main/scala/zio/redis/Command.scala
mijicd: I'll try to keep the interpretation here, and I'm quite flexible in terms of environment / layer dependencies at this moment. My plan was to get the rough API, get the "feel" of it, and tackle the interpretation in follow-up.

src/main/scala/zio/redis/Command.scala
mijicd: To avoid returning meaningless responses, I believe we'll have to tackle pipelining separately. It might be long down the road, but I'll definitely open it up as an issue as soon as we have a working prototype.

joeyfeldberg: Another alternative (although I'm not suggesting that it's good in any way) is to return an `Id[Out] = Out` or an `Queued[Out] = Unit`, meaning that the client with implement the API with Id and the Pipeline client with Queued.
Is this ugly?

Honestly the best solution is to pipeline everything like the Haskell client, but I'm not sure this can be done using ZIO's Monad.

mijicd: If you're talking about hedis, I know what you mean. That might be nice option.

jdegoes:
```scala
trait Queued[-R, +E, +A] { 
  // map, zip, zipLeft, zipRight, etc.

  def run: ZIO[R, E, A]
}
```

Then you can form composite commands like:

```scala
(echo("foo") *> ping *> del("foo")).execute
```

which will be fully pipelined.

Hedis uses a terrible hack to achieve its magic "through" bind. We can't do that in Scala, but even if we could, I'd want to avoid that.

src/main/scala/zio/redis/Command.scala
jdegoes: Probably `run` will look like this:

```scala
def run(in: In): Queued[RedisClient, RedisError, Out]
```

Actually, it occurs to me instead of `Queued`, we can probably use `ZQuery`, if we're willing to allow `flatMap` at the cost of breaking a pipeline. /cc @adamgfraser

src/main/scala/zio/redis/package.scala
```diff
@@ -0,0 +1,159 @@
+package zio
+
+package object redis {
+
+  import Command.Input._
+  import Command.Output._
+
+  object connection {
```
jdegoes: On second thought, let's make these traits, which are extended by these objects. That way we can have a `all` package, so users can do, `import redis.all._` or something.

src/main/scala/zio/redis/Command.scala
adamgfraser: Great! I have been delaying a bit on doing this just because there hasn't been other demand outside of Caliban to date and we have had other priorities with the new environment encoding and 1.0 release, but if there is interest here I can extract ASAP so it is something you can depend on.

mijicd: @adamgfraser Nice! Let's sync over discord and work on this as soon as I got this merged and you have some spare cycles.

src/main/scala/zio/redis/package.scala
mijicd: Makes sense.

mijicd:
@jdegoes Thanks for reviewing! I'll apply the object-trait conversion, and add the sorted sets API, then I'm merging this one.

src/main/scala/zio/redis/options/zadd.scala
```diff
@@ -0,0 +1,16 @@
...
+
+  object Updates {
+    case object XX extends Updates
+    case object NX extends Updates
+  }
+
+  case object CH
+  case object INCR
+
+  type CH   = CH.type
```
jdegoes: I like and recommend this pattern. 👍

jdegoes: There's a slight improvement on it:

```scala
sealed trait CH
val CH: CH = new CH
```

This way the value `CH` has type `CH`.

src/main/scala/zio/redis/api/SortedSets.scala
```diff
@@ -0,0 +1,38 @@
+package zio.redis.api
```
jdegoes: I might have a package object in `zio.redis.api` with:

```scala
package object api {
  val sortedsets = new SortedSets {}
  val sets = new Sets {}
}
```

For more "ala carte" imports.

Issue #14, opened 2020-03-06
Add missing sorted sets commands


mijicd:
Sorted sets API currently doesn't support the following:

- [x] `ZINTERSTORE`
- [x]  `ZLEXCOUNT`
- [x]  `ZPOPMAX`
- [x]  `ZPOPMIN`
- [x]  `ZRANGEBYLEX`
- [x]  `ZREVRANGEBYLEX`
- [x]  `ZRANGEBYSCORE`
- [x]  `ZREMRANGEBYLEX`
- [x]  `ZREMRANGEBYRANK`
- [x]  `ZREMRANGEBYSCORE`
- [x]  `ZREVRANGE`
- [x]  `ZREVRANGEBYSCORE`
- [x]  `ZREVRANK`
- [x]  `ZUNIONSTORE`

As there are quite a few of them, please raise your hand for the ones you're tackling :).

landlockedsurfer:
I will start working on this issue.

mijicd:
Closed via #17

PR #17, opened 2020-03-23
Add missing sorted sets commands

landlockedsurfer:
Some notes on this pull request for issue #14 :

* `ZUNIONSTORE` and `ZINTERSTORE`: I did not find a way to represent `list of simple keys | list of weighted keys`.  
In the current version both simple and weighted keys can still be mixed which is not allowed.
* Looks like there are three types of ranges used in sorted sets queries,
which can be used in both regular search,reversed search and removal queries.
At the moment those
```
  case object ZRangeWithScoresInput extends Input[options.zrange.WithScores]
  case object ZRangeByLexRangeInput extends Input[options.zrangebylex.Range]
  case object ZRangeByLexLimitInput extends Input[options.zrangebylex.Limit]
  case object ZRangeByScoreRangeInput extends Input[options.zrangebyscore.Range]
  case object ZRangeByScoreLimitInput extends Input[options.zrangebyscore.Limit]
  case object ZRangeByScoreWithScoresInput extends Input[options.zrangebyscore.WithScores]
```
are reused for the `zrev...` and `zrem...` variants.
It might be a useful to call the range types more explicitly `RankRange`,`LexRange`,`ScoreRange` instead of just `Range`.
* I think we might have to convert `WithScores` to an object since it needs to instantiable.
```
revrange("abc",Range(1,2),new WithScores {})  // won't work... WithScores is sealed, see also zrange options
```


src/main/scala/zio/redis/api/SortedSets.scala
```diff
@@ -10,29 +10,103 @@ trait SortedSets {
...
 
   final val zcard   = Command("ZCARD", StringInput, LongOutput)
   final val zcount  = Command("ZCOUNT", Tuple2(StringInput, RangeInput), LongOutput)
   final val zincrby = Command("ZINCRBY", Tuple3(StringInput, LongInput, ByteInput), LongOutput)
+
+  final val zlexcount = Command("ZLEXCOUNT", Tuple2(StringInput, ZLexCountInput), LongOutput)
+
+  final val zinterstore = Command(
+    "ZINTERSTORE",
+    Tuple3(
```
mijicd: You can make it a `Tuple4` and separate keys from weights. Later on, one syntax class needs to be added that has curried parameters to support parameters after non-empty list.

src/main/scala/zio/redis/options/zinterstore.scala
```diff
@@ -0,0 +1,17 @@
+package zio.redis.options
+
+object zinterstore {
+
+  sealed trait Key
+  object Key {
+    case class SimpleKey(key: String)                   extends Key
```
mijicd: Please make all case classes final.

Also, having in mind the comment above, you might end up without `Key`, but with 2 simple case classes.

src/main/scala/zio/redis/options/zrangebylex.scala
```diff
@@ -0,0 +1,22 @@
+package zio.redis.options
+
+object zrangebylex {
+
+  sealed trait Minimum
```
mijicd: The docs for `ZLEXCOUNT` states `The min and max arguments have the same meaning as described for `ZRANGEBYLEX`. So, instead of re-defining it for every command, I'd just leave this one, and use it as an input for other commands referring to it.

**Edit:** I realized I might have mislead you in our chat regarding flattening of the options package. Sorry for that. Let's not postpone that and flatten the options package immediately:

```scala
object options {
// all custom options
}
```

You can call the file `options.scala` and put it in the same package with `Input`. That would help reducing the noise, e.g. no need for limit inputs specific to each command.

src/main/scala/zio/redis/api/SortedSets.scala
```diff
@@ -10,29 +10,103 @@ trait SortedSets {
...
+    ),
+    StreamOutput
+  )
+
+  final val zrangebyscore = Command(
+    "ZRANGEBYSCORE",
+    Tuple4(
+      StringInput,
+      ZRangeByScoreRangeInput,
+      OptionalInput(ZRangeWithScoresInput),
```
mijicd: Regarding the comment in PR description: please make `WithScores` a case object. That should suffice. Alternatively, you can have a helper lazy val with instantiated `WithScores`.

src/main/scala/zio/redis/api/SortedSets.scala
```diff
@@ -10,29 +10,103 @@ trait SortedSets {
...
     "ZSCAN",
     Tuple4(LongInput, OptionalInput(MatchInput), OptionalInput(LongInput), OptionalInput(StringInput)),
     ScanOutput
   )
   final val zscore = Command("ZSCORE", Tuple2(StringInput, ByteInput), LongOutput)
+
+  final val zunionstore = Command(
+    "ZUNIONSTORE",
+    Tuple3(
+      StringInput,
```
mijicd: I think you're missing `numkeys`.

src/main/scala/zio/redis/api/SortedSets.scala
landlockedsurfer: The idea was to represent the fact that `ZINTERSTORE` and `ZUNIONSTORE` can only either have a non-empty list of keys or a non-empty list of key/weight pairs using the type system. I did away with the `numkeys` parameter, since I thought it would be given by the list itself. However as far as I understood... I misunderstood and you have something else in mind. :-)
I have splitted keys and weights in the next version.

src/main/scala/zio/redis/options/zinterstore.scala
landlockedsurfer: After reorganizing the modules into `shared`, `zset` etc. and making the case classes final in the `zset`trait I ran into warnings like:
```
[warn] /Users/manfred/src/github/zio-redis/src/main/scala/zio/redis/options/zset.scala:8:22: The outer reference in this type test cannot be checked at run time.
[warn]     final case class Open(value: String)   extends LexMinimum
```

src/main/scala/zio/redis/options/zrangebylex.scala
landlockedsurfer: As discussed in our chat the plan is to group the types for the commands in a similar fashion like it is done on the [Redis commands documentation page itself ](https://redis.io/commands).
Additional there will be a shared module to declare cross-cutting command options.

src/main/scala/zio/redis/api/SortedSets.scala
landlockedsurfer: I've used case objects for now.

src/main/scala/zio/redis/api/SortedSets.scala
landlockedsurfer: See my first comment.

landlockedsurfer:
Some comments on this pull request:
* Do you prefer SortedSet or ZSet? SortedSet is more verbose, but it might be a better fitting.
* Is it appropriate to type aliase the new objects? E.g.
```
case object WithScores
type WithScores = WithScores.type
```
* I could not make the case classes final due to warnings like
```
[warn] /Users/manfred/src/github/zio-redis/src/main/scala/zio/redis/options/zset.scala:8:22: The outer reference in this type test cannot be checked at run time.
[warn]     final case class Open(value: String)   extends LexMinimum```
```

src/main/scala/zio/redis/api/SortedSets.scala
mijicd: Ah, but they can both be present, see the example from command docs: `ZUNIONSTORE out 2 zset1 zset2 WEIGHTS 2 3` :).

src/main/scala/zio/redis/options/zinterstore.scala
mijicd: Oh, I hate this one. You should be able to seal it instead.

src/main/scala/zio/redis/api/SortedSets.scala
mijicd: You're right about `numkeys` though. That one's redis being "silly" :).

mijicd:
@landlockedsurfer It is appropriate to alias, we had that initially, and replaced it with sealed as it seemed nicer :).

I prefer `sortedset`. Prefix `z` is already heavily "overloaded" in ZIO universe :).

src/main/scala/zio/redis/api/SortedSets.scala
```diff
@@ -10,29 +10,107 @@ trait SortedSets {
...
   final val zcount  = Command("ZCOUNT", Tuple2(StringInput, RangeInput), LongOutput)
   final val zincrby = Command("ZINCRBY", Tuple3(StringInput, LongInput, ByteInput), LongOutput)
+
+  final val zlexcount = Command("ZLEXCOUNT", Tuple2(StringInput, ZSetLexRangeInput), LongOutput)
+
+  final val zinterstore = Command(
+    "ZINTERSTORE",
+    Tuple5(
+      StringInput,
+      LongInput,
```
mijicd: `numkeys` is actually very ugly parameter. I'd improve that it in subsequent issues as we need to ensure 2 lists are of the same length. For now we can drop `numkeys` and return immediately if lenghts are not equal.

src/main/scala/zio/redis/options/shared.scala
```diff
@@ -0,0 +1,5 @@
+package zio.redis.options
+
+trait shared extends zset {
```
mijicd: You can use capitalized names for these, they're all subsumed by `options` anyhow. Also, no need to extend `zset` here (done in `options`).

src/main/scala/zio/redis/options/zset.scala
```diff
@@ -0,0 +1,62 @@
+package zio.redis.options
+
+trait zset {
```
mijicd: Rename to `SortedSets` to mimic the API.

src/main/scala/zio/redis/api/SortedSets.scala
landlockedsurfer: As discussed we will settle on this version for `ZUNIONSTORE` (and `ZINTERSTORE`) for now:
```
  final val zunionstore = Command(
    "ZUNIONSTORE",
    Tuple4(
      StringInput,
      NonEmptyList(StringInput),
      OptionalInput(NonEmptyList(DoubleInput)),
      OptionalInput(SortedSetAggregateInput)
    ),
    LongOutput
  )
```

src/main/scala/zio/redis/options/zinterstore.scala
landlockedsurfer: All sealed up! I.e. replaced `case class` with `sealed case class` in `zio/redis/options/SortedSets.scala`.

src/main/scala/zio/redis/options/shared.scala
landlockedsurfer: As discussed in our chat, we'll get rid of `shared` all together.

mijicd:
Please reformat the sources and let's merge. Nice work!

manfredweber:
Missed that one. :-) Ran the `fmt` task again and pushed my branch again. Hope formatting is fine now.

PR #19, opened 2020-03-29
Add geo commands

### LINKED ISSUE #5: Represent geo commands
https://redis.io/commands#geo

paulpdaniels:
Closes #5

### DIFF
--- a/src/main/scala/zio/redis/Input.scala
+++ b/src/main/scala/zio/redis/Input.scala
@@ -28,6 +28,15 @@ object Input {
   case object SortedSetWithScoresInput  extends Input[options.WithScores]
   case object SortedSetLimitInput       extends Input[options.Limit]
   case object SortedSetScoreRangeInput  extends Input[options.ScoreRange]
+  case object GeoCountInput             extends Input[options.Count]
+  case object GeoLongLatInput           extends Input[options.LongLat]
+  case object GeoOrderInput             extends Input[options.Order]
+  case object GeoRadiusUnitInput        extends Input[options.RadiusUnit]
+  case object GeoStoreInput             extends Input[options.Store]
+  case object GeoStoreDistInput         extends Input[options.StoreDist]
+  case object GeoWithCoordInput         extends Input[options.WithCoord]
+  case object GeoWithDistInput          extends Input[options.WithDist]
+  case object GeoWithHashInput          extends Input[options.WithHash]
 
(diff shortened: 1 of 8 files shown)


src/main/scala/zio/redis/api/Geo.scala
```diff
@@ -0,0 +1,84 @@
...
+  lazy val geoDist =
+    Command(
+      "GEODIST",
+      Tuple4(
+        StringInput,
+        StringInput,
+        StringInput,
+        OptionalInput(GeoRadiusUnitInput)
+      ),
+      StringOutput
```
mijicd: I'd represent it as double, optional one according to the specs.

src/main/scala/zio/redis/api/Geo.scala
```diff
@@ -0,0 +1,84 @@
...
+    )
+
+  lazy val geoHash =
+    Command(
+      "GEOHASH",
+      NonEmptyList(StringInput),
```
mijicd: I'd keep key separate, e.g. `Tuple2(StringInput, NonEmptyList(StringInput))`.

src/main/scala/zio/redis/Input.scala
```diff
@@ -52,5 +61,19 @@ object Input {
     _6: Input[F]
   ) extends Input[(A, B, C, D, E, F)]
 
+  final case class Tuple11[-A, -B, -C, -D, -E, -F, -G, -H, -I, -J, -K](
```
mijicd: Make sure to add syntax for this to Command's companion. Also, please delete `Tuple6` as it doesn't seem to be used.

src/main/scala/zio/redis/api/Geo.scala
paulpdaniels: We were missing both `OptionalOutput` and `DoubleOutput` so I added them, let me know if that is not what you meant.

src/main/scala/zio/redis/api/Geo.scala
paulpdaniels: Done

src/main/scala/zio/redis/Input.scala
paulpdaniels: Done

mijicd:
@paulpdaniels fix the imports, looks good otherwise!

Issue #11, opened 2020-03-06
Add missing strings commands


mijicd:
Strings API currently doesn't support:
- [x] `BITFIELD`
- [x] `BITOP`
- [x] `BITPOS`
- [x] `SET`

Please raise your hand for the ones you're taking :).

landlockedsurfer:
I will start working on this one.

mijicd:
Closed in #21

PR #21, opened 2020-03-29
Add missing strings commands

landlockedsurfer:
First version of my pull request to close issue #11.

Here are points that still need to be clarification:
* BigInt for `BITFIELD`'s value parameter?
* `BitPosRange`: Crated new range type instead of using Range here since the upper bound is optional.
```
case class BitPosRange(start:Long,end:Option[Long])
``` 
* Decided to put `Updates` into package object:
  * Is this a good idea? Or do you prefer keeping two versions for the `String` (`SET`) and `SortedSet` (`ZADD`) commands
  * Got warning about not putting trait directly into package object; To get rid of it, we could pick up the idea again of having a `shared` trait.
  ```
  [warn] /Users/manfred/src/github/zio-redis/src/main/scala/zio/redis/options/package.scala:6:16: it is not recommended to define classes/objects inside of package objects.
  [warn] If possible, define trait Updates in package options instead.
  [warn]   sealed trait Updates
  [warn]                ^
  ```
  * If we shared `Updates`, would we want to use a different name?
* `SET` can return either a OK or nil
Example redis-cli session:
```
> set abc 2 EX 2 NX
OK
> set abc 2 EX 2 XX
(nil)
```
Thus have used `OptionalOutput` introduced by @paulpdaniels commit.


src/main/scala/zio/redis/options/Strings.scala
```diff
@@ -0,0 +1,39 @@
+package zio.redis.options
+
+trait Strings {
+
+  sealed trait BitFieldType
+  object BitFieldType {
+    sealed case class UnsignedInt(size: Int) extends BitFieldType
```
mijicd, line 7: This isn't completely safe as it allows creation of invalid values, e.g. `UnsignedInt(1000)`. Instead, I'd use smart constructors here, we'll work on usability later.

src/main/scala/zio/redis/options/Strings.scala
```diff
@@ -0,0 +1,39 @@
...
+    case object AND extends BitOperation
+    case object OR  extends BitOperation
+    case object XOR extends BitOperation
+  }
+
+  case class BitPosRange(start: Long, end: Option[Long])
+
+  sealed trait Expiration
+  object Expiration {
+    sealed case class EX(seconds: Long)
```
mijicd, line 33: `extends Expiration`?

src/main/scala/zio/redis/options/Strings.scala
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  sealed trait BitOperation
+  object BitOperation {
+    case object AND extends BitOperation
+    case object OR  extends BitOperation
+    case object XOR extends BitOperation
+  }
+
+  case class BitPosRange(start: Long, end: Option[Long])
```
mijicd, line 29: We can evolve this into general solution, something like:

```scala
sealed trait Range

object Range {
  final case class LeftClosed(a: Long) extends Range
  final case class LeftClosedRightOpen(a: Long, b: Long) extends Range
  final case class Closed(a: Long, b: Long) extends Range
  // ...
}
```

The names can be improved as they're math-inspired, and we probably need only combinations given above. On top of these, we can add some helpers to create them.

src/main/scala/zio/redis/api/Strings.scala
```diff
@@ -20,9 +42,20 @@ trait Strings {
...
-  final val setbit      = Command("SETBIT", Tuple3(StringInput, LongInput, ByteInput), ByteOutput)
-  final val setex       = Command("SETEX", Tuple3(StringInput, DurationInput, ByteInput), UnitOutput)
-  final val setnx       = Command("SETNX", Tuple2(StringInput, ByteInput), BoolOutput)
-  final val setrange    = Command("SETRANGE", Tuple3(StringInput, LongInput, ByteInput), LongOutput)
-  final val strlen      = Command("STRLEN", StringInput, LongOutput)
+  final val set = Command(
+    "SET",
+    Tuple5(
+      StringInput,
+      StringInput,
```
mijicd, line 49: `ByteInput` is better suited here.

src/main/scala/zio/redis/api/Strings.scala
```diff
@@ -20,9 +42,20 @@ trait Strings {
...
+      OptionalInput(StringExpirationInput),
+      OptionalInput(UpdatesInput),
+      OptionalInput(StringKeepTimeToLiveInput)
```
mijicd, line 52: This one (and similar) can be modeled by boolean input with some default. Leave it for now, we'll see how the API looks later on!

src/main/scala/zio/redis/options/Strings.scala
landlockedsurfer: Below a complete list of all possible combinations between open/closed and bounded/unbounded intervals.

| Interval | Type |
| --------- | ------------- |
| (a,b) | `Open` |
| [a,b] | `Closed` |
| (a,b] | `LeftOpenRightClosed` | 
| [a,b) | `LeftClosedRightOpen` |
| (a,+∞) | `LeftOpenRightUnbounded` |
| [a,+∞) | `LeftClosedRightUnbounded` |
| (−∞,b) | `LeftUnboundedRightOpen` |
| (−∞,b] | `LeftUnboundedRightClosed` |
| (−∞,+∞) | `Unbounded` |

```
sealed trait Range[T]
object Range {
    final case class Open[T](lo:T,hi:T) extends Range[T]
    final case class Closed[T](lo:T,hi:T) extends Range[T]
    final case class LeftOpenRightClosed[T](lo:T,hi:T) extends Range[T]
    final case class LeftClosedRightOpen[T](lo:T,hi:T) extends Range[T]
    final case class LeftOpenRightUnbounded[T](lo:T) extends Range[T]
    final case class LeftClosedRightUnbounded[T](lo:T) extends Range[T]
    final case class LeftUnboundedRightOpen[T](hi:T) extends Range[T]
    final case class LeftUnboundedRightClosed[T](hi:T) extends Range[T]
    final case object Unbounded extends Range[Nothing]
}
```

## Range Usage

Here is an overview of all commands we have made available so far grouped by range types.

### Commands accepting `Closed`

`LTRIM`,`LRANGE`,`ZRANGE`,`ZREVRANGE`,`ZREMRANGEBYRANK`

For each command negative indices means counted from the end of range.

Possible range input type for those commands: `Input[options.Range.Closed[Long]]`.

### Commands accepting `Closed` or `LeftClosedRightUnbounded`

`BITPOS`, `BITCOUNT`

For each command the range is optional.

Possible range input type for those commands:
`Input[Either[options.Range.Closed[Long],options.Range.LeftClosedRightUnbounded[Long]]]`.

### Commands accepting any Interval

`ZCOUNT` (using +inf,-inf for infinite intervals)
`ZRANGEBYSCORE`,`ZREVRANGEBYSCORE`,`ZREMRANGEBYSCORE` (using +inf,-inf for infinite intervals)
`ZLEXCOUNT`,`ZRANGEBYLEX`,`ZREVRANGEBYLEX`,`ZREMRANGEBYLEX` (using +,- for infinite intervals)

Possible range input type for score set commands:
`Input[options.Range[Double]`

Possible range input type for lexicographical set commands:
`Input[options.Range[String]`

src/main/scala/zio/redis/options/Strings.scala
landlockedsurfer: Could you please elaborate on this a little more? Using smart constructors would require error handling, which we have not considered so far.

src/main/scala/zio/redis/options/Strings.scala
landlockedsurfer: Ups! Should now be fixed. :-)

src/main/scala/zio/redis/api/Strings.scala
landlockedsurfer: Done!

src/main/scala/zio/redis/options/Strings.scala
mijicd: That's also true, and perhaps out of scope. We can note this one down and let it "leak" inside the error channel of response.

src/main/scala/zio/redis/options/Strings.scala
mijicd: Nice work! I propose that you open up a separate PR and introduce ranges. In case we don't need some of the options, you can cut down the hierarchy.

PR #51, opened 2020-05-26
add migrate, randomkey, restore and wait keys

mikail-khan:


src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -29,4 +57,5 @@ trait Keys {
   final val ttl    = RedisCommand("TTL", StringInput, DurationOutput)
   final val typeOf = RedisCommand("TYPE", StringInput, StringOutput)
   final val unlink = RedisCommand("UNLINK", NonEmptyList(StringInput), LongOutput)
+  final val wait_  = RedisCommand("WAIT", Tuple2(LongInput, LongInput), LongOutput)
```
mikail-khan, line 60: appended an underscore until we know what to do with this

src/main/scala/zio/redis/options/Shared.scala
```diff
@@ -1,6 +1,8 @@
 package zio.redis.options
 
 trait Shared {
+  sealed case class Limit(offset: Int, count: Int)
```
mikail-khan, line 4: shared by `SortedSets` and `Keys`

mijicd: It doesn't seem to be used in keys though.

mikail-khan: I made that change to implement the `sort` command but then realised that `sort` can't be introduced yet until the current design can accommodate it. I can move it back to `SortedSets` if `sort` is not viable.

mijicd: Already did it on interpreter.

PR #72, opened 2020-06-24
Boost performance

mijicd:
### Summary

- Drop string builder.
- Fuse effects and complete promise unsafely.
- Switch to blocking thread pool.
- Represent errors as exceptions and dropped eithers.

### DIFF
--- a/redis/src/main/scala/zio/redis/Interpreter.scala
+++ b/redis/src/main/scala/zio/redis/Interpreter.scala
@@ -62,14 +62,18 @@ trait Interpreter {
       private val executeNext: UIO[Any] =
         pendingRequests.take.flatMap { request =>
           val exchange =
-            IO {
-              unsafeSend(request.command)
-              unsafeReceive()
+            IO.effect {
+              try {
+                unsafeSend(request.command)
+                unsafeReceive()
+              } catch {
+                case t: Throwable => throw RedisError.ProtocolError(t.getMessage)
+              } finally {
+                readinessFlag.set(true)
+              }
             }
 
-          val completer = exchange.catchAll(RedisError.make).ensuring(UIO(readinessFlag.set(true)))
-
-          request.result.completeWith(completer)
+          request.result.completeWith(exchange.refineToOrDie[RedisError])
         }
 
       private def unsafeReceive(): String = {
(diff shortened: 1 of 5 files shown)


redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -60,35 +62,39 @@ trait Interpreter {
...
+                unsafeReceive()
+              } catch {
+                case t: Throwable => throw RedisError.ProtocolError(t.getMessage)
+              } finally readinessFlag.set(true)
             }
 
-          val completer = exchange.catchAll(RedisError.make).ensuring(UIO(readinessFlag.set(true)))
-
-          request.result.completeWith(completer)
+          request.result.unsafeDone(exchange.refineToOrDie[RedisError])
```
jdegoes, line 76: Technically we are doing an effect inside `map`, which is a no-no, but I think this one is safe to do and gives us a performance boost, so 👍

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -60,35 +62,39 @@ trait Interpreter {
...
         }
 
       private def unsafeReceive(): String = {
-        val sb        = new StringBuilder
+        val cb        = ChunkBuilder.make[Byte]()
```
jdegoes, line 80: Actually, I suggest: `ArrayBuilder[Chunk[Byte]]` (or maybe `ArrayBuilder[Array[Byte]]`. The `++=` on `ChunkBuilder` will end up boxing `Byte` (I think).

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -60,35 +62,39 @@ trait Interpreter {
...
         // TODO: handle -1
         while (readBytes == 0) {
           channel.read(response)
           response.flip()
 
-          readBytes = response.limit()
+          readBytes = response.remaining()
 
-          sb ++= new String(Arrays.copyOf(response.array(), readBytes), UTF_8)
+          cb ++= Chunk.fromByteBuffer(response)
```
jdegoes, line 90: I don't know if `Chunk` will copy the byte buffer, so this may not be safe. I think if you call `.materialize`, it will make an array copy. We should verify.

mijicd: It should be:

```scala
    val dest = Array.ofDim[Byte](buffer.remaining())
    val pos  = buffer.position()
    buffer.get(dest)
    buffer.position(pos)
    Chunk.fromArray(dest)
```

This would be problematic if buffer isn't flipped, otherwise, I don't think it'll make any problems.

PR #74, opened 2020-06-25
Test keys API

### LINKED ISSUE #58: Test keys API
Expand the [ApiSpec](https://github.com/zio/zio-redis/blob/master/src/test/scala/zio/redis/ApiSpec.scala) with remaining keys commands.

aleksandarskrbic:
Closes #58 

**Summary:**
1. Implement test for keys API
2. Add SimpleStringOutput, needed for type command response decoding

**Notes:**
1. Inspect DUMP command
2. KEYS command sometimes works sometimes not, I realized that it fails when there are 24 or more keys


redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -161,6 +161,14 @@ object Output {
       Either.cond(text == "+OK\r\n", (), ProtocolError(s"$text isn't unit."))
   }
 
+  case object SimpleStringOutput extends Output[String] {
```
mijicd, line 164: When I think more about this, I think we should do the following:

- Rename StringOutput to MultiStringOutput.
- Rename this to StringOutput.
- Create an ADT to represent possible redis types.
- Create a TypeOutput.

While the newly created StringOutput won't be in use, we might keep it until we verify there are no more "mistypes".

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -1,8 +1,13 @@
 package zio.redis
 
+import java.time.Instant
 import java.util.UUID
 
+import zio.test.testM
+import java.util.concurrent.TimeUnit
```
mijicd, line 7: Move it to "java group". I'll open an issue for adding scalafix rules.

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -39,10 +44,170 @@ object ApiSpec extends BaseSpec {
...
+        },
+        testM("find all keys matching pattern") {
+          for {
+            value    <- uuid
+            _        <- set("custom_key_1", value, None, None, None)
+            _        <- set("custom_key_2", value, None, None, None)
+            response <- keys("*custom*")
+          } yield assert(response.length)(equalTo(2))
+        },
+        testM("pExpire followed by pTtl") {
```
mijicd, line 64: I would group these in a suite called "expiration".

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -39,10 +44,170 @@ object ApiSpec extends BaseSpec {
...
+            timeout  <- instantForNextMillis(5000L)
+            _        <- set(key, value, None, None, None)
+            exp      <- expireAt(key, timeout)
+            response <- get(key)
+            ttl      <- ttl(key)
+          } yield assert(exp)(isTrue) &&
+            assert(response)(isSome) &&
+            assert(ttl.toMillis)(isLessThan(timeout.toEpochMilli))
+        },
+        testM("rename existing key") {
```
mijicd, line 112: Group renames in a suite called "renaming".

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -39,10 +44,170 @@ object ApiSpec extends BaseSpec {
...
+            renamed <- rename(key, newKey).either
+          } yield assert(renamed)(isLeft)
+        },
+        testM("rename key to newkey if newkey does not yet exist") {
+          for {
+            key     <- uuid
+            newKey  <- uuid
+            value   <- uuid
+            _       <- set(key, value, None, None, None)
+            renamed <- renameNx(key, newKey)
```
mijicd, line 135: Please add the test for renaming existing key as well.

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -39,10 +44,170 @@ object ApiSpec extends BaseSpec {
...
+            key1    <- uuid
+            value1  <- uuid
+            _       <- set(key1, value1, None, None, None)
+            key2    <- uuid
+            value2  <- uuid
+            _       <- set(key2, value2, None, None, None)
+            touched <- touch(key1, List(key2))
+          } yield assert(touched)(equalTo(2L))
+        },
+        testM("set followed by scan") {
```
mijicd, line 165: No need to include set in a test description, i.e. treat set as necessary evil :).

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -39,10 +44,170 @@ object ApiSpec extends BaseSpec {
...
+          } yield assert(stringType)(equalTo("string")) && assert(listType)(equalTo("list"))
+        },
+        testM("dump followed by restore") {
+          for {
+            key      <- uuid
+            value    <- uuid
+            _        <- set(key, value, None, None, None)
+            dumped   <- dump(key)
+            _        <- del(key, Nil)
+            restore  <- restore((key, 0L, dumped, None, None, None, None)).either
```
mijicd, line 202: Please define `Arg7` syntax class in `RedisCommand`, no need to pass tuple explicitly.

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -39,10 +44,170 @@ object ApiSpec extends BaseSpec {
...
+          } yield assert(restore)(isRight) && assert(restored)(isSome(equalTo(value)))
         }
       )
     ).provideCustomLayerShared(Executor)
 
-  private val Executor = RedisExecutor.live("127.0.0.1", 6379).orDie
-  private val uuid     = UIO(UUID.randomUUID().toString)
+  private val Executor                                          = RedisExecutor.live("127.0.0.1", 6379).orDie
+  private val uuid                                              = UIO(UUID.randomUUID().toString)
+  private def durationInMillis(seconds: Long): UIO[Duration]    = UIO(Duration(seconds, TimeUnit.MILLISECONDS))
```
mijicd, line 211: You can remove the need for this by `import zio.duration._`. Every call to `durationInMillis` becomes `something.millis`.

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -39,10 +44,170 @@ object ApiSpec extends BaseSpec {
...
+  private def instantForNextMillis(seconds: Long): UIO[Instant] = UIO(Instant.now().plusMillis(seconds))
```
mijicd, line 212: Rename this to `instantOf`.

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -39,10 +44,170 @@ object ApiSpec extends BaseSpec {
...
+            deleted <- del(key, Nil)
+          } yield assert(deleted)(equalTo(1L))
+        },
+        testM("find all keys matching pattern") {
+          for {
+            value    <- uuid
+            _        <- set("custom_key_1", value, None, None, None)
+            _        <- set("custom_key_2", value, None, None, None)
+            response <- keys("*custom*")
+          } yield assert(response.length)(equalTo(2))
```
mijicd, line 62: I'd make an assertion on list elements.

redis/src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -4,9 +4,19 @@ import zio.redis.Input._
...
+sealed trait Type
+
+object Type {
+  case object String    extends Type
+  case object List      extends Type
+  case object Set       extends Type
+  case object SortedSet extends Type
+  case object Hash      extends Type
+}
+
```
mijicd, line 16: Move it to keys options.

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -2,6 +2,7 @@ package zio.redis
 
 import zio.Chunk
 import zio.duration._
+import zio.redis.api.Type
```
mijicd, line 5: There should be no need for this as it is imported implicitly through package object inheritance.

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -39,10 +45,204 @@ object ApiSpec extends BaseSpec {
...
+              _     <- set(key, value, None, None, None)
+              exp   <- pExpire(key, 100.millis)
+              ttl   <- pTtl(key).either
+            } yield assert(exp)(isTrue) && assert(ttl)(isRight)
+          },
+          testM("pExpireAt followed by get and pTtl") {
+            for {
+              key      <- uuid
+              value    <- uuid
+              timeout  <- instantOf(100L)
```
mijicd, line 129: This isn't a timeout, rather an exact instant of when this will expire.

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -39,10 +45,204 @@ object ApiSpec extends BaseSpec {
...
+              _        <- set(key, value, None, None, None)
+              exp      <- pExpireAt(key, timeout)
+              response <- get(key)
+              ttl      <- pTtl(key)
```
mijicd, line 133: Note that you're guarding the similar call with either in a test above. To mitigate the flakiness, I'd rather do this with 2 get calls and sleep in between. Assertion would be that the first get returns something, while the second returns none. Please update all other tests as well.

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -39,10 +44,204 @@ object ApiSpec extends BaseSpec {
...
+          }
+        ),
+        suite("types")(
+          testM("string type") {
+            for {
+              key    <- uuid
+              value  <- uuid
+              _      <- set(key, value, None, None, None)
+              string <- typeOf(key)
+            } yield assert(string)(isSubtype[RedisType.String.type](anything))
```
mijicd, line 205: equalTo?

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -39,10 +44,204 @@ object ApiSpec extends BaseSpec {
...
+          },
+          testM("expireAt followed by get and ttl") {
+            for {
+              key       <- uuid
+              value     <- uuid
+              expiresAt <- instantOf(2000)
+              _         <- set(key, value, None, None, None)
+              _         <- expireAt(key, expiresAt)
+              response1 <- exists(key, Nil)
+              ttl       <- ttl(key)
```
mijicd, line 156: I don't like this test, as it mixes concerns. Checking the ttl should be done separately. If you can do it quickly, go for it, otherwise, let's do it as follow up.

PR #91, opened 2020-07-29
Test sets API

### LINKED ISSUE #63: Test sets API
Expand the [ApiSpec](https://github.com/zio/zio-redis/blob/master/src/test/scala/zio/redis/ApiSpec.scala) with sets commands.

anovakovic01:
This PR resolves #63.

**Summary:** Implement tests for the sets API.

**Notes:**
1. Looks like `SSCAN` command is missing `key` parameter.
2. Some of the tests for `SPOP` are failing for an unknown reason. From what I can tell, something breaks when parsing output.
3. Tests for `SRANDMEMBER` are failing when there is no count, i.e. output consists of only one element.

### DIFF
--- a/redis/src/test/scala/zio/redis/SetsSpec.scala
+++ b/redis/src/test/scala/zio/redis/SetsSpec.scala
@@ -383,6 +383,29 @@ trait SetsSpec extends BaseSpec {
             isMember <- sIsMember(key, "a").either
           } yield assert(isMember)(isLeft(isSubtype[WrongType](anything)))
         }
+      ),
+      suite("members")(
+        testM("sMembers non-empty set") {
+          for {
+            key <- uuid
+            _ <- sAdd(key)("a", "b", "c")
+            members <- sMembers(key)
+          } yield assert(members)(hasSameElements(Chunk("a", "b", "c")))
+        },
+        testM("sMembers empty set") {
+          for {
+            key <- uuid
+            members <- sMembers(key)
+          } yield assert(members)(isEmpty)
+        },
+        testM("sMembers when not set") {
+          for {
+            key <- uuid
+            value <- uuid
+            _ <- set(key, value, None, None, None)
+            members <- sMembers(key).either
+          } yield assert(members)(isLeft(isSubtype[WrongType](anything)))
+        }
       )
     )
 }
--- a/redis/src/test/scala/zio/redis/SetsSpec.scala
+++ b/redis/src/test/scala/zio/redis/SetsSpec.scala
@@ -8,8 +8,8 @@ import zio.test.Assertion._
 trait SetsSpec extends BaseSpec {
   val setsSuite =
     suite("sets")(
-      suite("add")(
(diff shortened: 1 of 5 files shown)


mijicd:
@anovakovic01 regarding the remarks you made:

1. Good catch, answered in the ticket.
2. Please check the data returned and compare it to the parsing output. Neat way to do this is by sending command via telnet to the local redis instance you have.
3. `SRANDMEMBER` isn't complete. As you see in the code, there's a comment left about multiple possible outputs. That said, we have 2 approaches here:

- Remove command and tests, and add a ticket to support it (I'm kinda leaning toward this one).
- Fix it in here (you can try and see how big of an effort this is).

anovakovic01:
@mijicd Thanks for the review, will fix it. Regarding `sRandMember`, maybe we could leave only tests for when the count is present, and add tests without count later?

mijicd:
@anovakovic01 that's also an option, but make sure to open a ticket to fix the other part.

redis/src/test/scala/zio/redis/SetsSpec.scala
```diff
@@ -0,0 +1,774 @@
...
+
+import zio.Chunk
+import zio.redis.RedisError.WrongType
+import zio.test._
+import zio.test.Assertion._
+
+trait SetsSpec extends BaseSpec {
+  val setsSuite =
+    suite("sets")(
+      suite("sAdd")(
```
mijicd, line 11: Add one more: adding multiple keys.

redis/src/test/scala/zio/redis/SetsSpec.scala
```diff
@@ -0,0 +1,774 @@
...
+            first  <- uuid
+            second <- uuid
+            _      <- sAdd(first)("a", "b", "c", "d")
+            _      <- sAdd(second)("a", "c")
+            diff   <- sDiff(first, List(second))
+          } yield assert(diff)(hasSameElements(Chunk("b", "d")))
+        },
+        testM("non-empty set and empty set") {
+          for {
+            first  <- uuid
```
mijicd, line 73: Nitpick: I'd name them `empty` and `nonEmpty`.

redis/src/test/scala/zio/redis/SetsSpec.scala
anovakovic01: Good catch, missed that one.

anovakovic01:
@mijicd I've addressed all of the remarks. One of the CI jobs is failing, can you rerun it?

mijicd:
@anovakovic01 Retriggered, and it is passing. I don't see the test for adding multiple keys.

anovakovic01:
@mijicd It's [here](https://github.com/zio/zio-redis/pull/91/files#diff-34bde252a2520bbabc09e17ea03befd5R32). If the test case label is confusing, please let me know, I can change it.

mijicd:
@anovakovic01 ah cool, I was viewing the stale diff.

PR #100, opened 2020-08-06
Test sorted sets API

### LINKED ISSUE #65: Test sorted sets API
Expand the [ApiSpec](https://github.com/zio/zio-redis/blob/master/src/test/scala/zio/redis/ApiSpec.scala) with sorted sets commands.

### LINKED ISSUE #97: Problem in sorted sets with: zInterStore, zUnionStore, zScan
During creating of unit tests for sorted sets: https://github.com/zio/zio-redis/issues/65
I found out that there are some problems with those 3 commands.
IMO looks like they are not properly "rendered" into redis commands.

I am creating this ticket and will work on it.

maciejbak85:
Resolves #65, resolves #97

Fixed: zInterStore, zUnionStore, zScan
zScore and a problem with INCR in zAdd ( https://github.com/zio/zio-redis/issues/97 )

### DIFF
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -20,7 +20,7 @@ object Input {
   }
 
   case object AggregateInput extends Input[Aggregate] {
-    def encode(data: Aggregate): Chunk[String] = Chunk.single(wrap(data.stringify))
+    def encode(data: Aggregate): Chunk[String] = Chunk(wrap("AGGREGATE"), wrap(data.stringify))
   }
 
   case object AuthInput extends Input[Auth] {
@@ -139,6 +139,13 @@ object Input {
       (data._1 :: data._2).foldLeft(Chunk.empty: Chunk[String])((acc, a) => acc ++ input.encode(a))
   }
 
+  final case class NonEmptyListPrecededByCount[-A](input: Input[A]) extends Input[(A, List[A])] {
+    def encode(data: (A, List[A])): Chunk[String] =
+      (data._1 :: data._2).foldLeft(Chunk.single(wrap((1 + data._2.length).toString)): Chunk[String])((acc, a) =>
+        acc ++ input.encode(a)
+      )
+  }
+
   case object OrderInput extends Input[Order] {
     def encode(data: Order): Chunk[String] = Chunk.single(wrap(data.stringify))
   }
@@ -185,6 +192,13 @@ object Input {
 
   case object TimeMillisecondsInput extends Input[Instant] {
     def encode(data: Instant): Chunk[String] = Chunk.single(wrap(data.toEpochMilli.toString))
+  }
+
+  case object WeightsInput extends Input[Weights] {
+    def encode(data: Weights): Chunk[String] =
+      data.weights.foldLeft(Chunk.single(wrap("WEIGHTS")): Chunk[String])((acc, a) =>
+        acc ++ Chunk.single(wrap(a.toString))
+      )
   }
 
   final case class Tuple2[-A, -B](_1: Input[A], _2: Input[B]) extends Input[(A, B)] {
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -139,13 +139,6 @@ object Input {
       (data._1 :: data._2).foldLeft(Chunk.empty: Chunk[String])((acc, a) => acc ++ input.encode(a))
   }
 
-  final case class NonEmptyListPrecededByCount[-A](input: Input[A]) extends Input[(A, List[A])] {
-    def encode(data: (A, List[A])): Chunk[String] =
-      (data._1 :: data._2).foldLeft(Chunk.single(wrap((1 + data._2.length).toString)): Chunk[String])((acc, a) =>
-        acc ++ input.encode(a)
-      )
-  }
-
   case object OrderInput extends Input[Order] {
     def encode(data: Order): Chunk[String] = Chunk.single(wrap(data.stringify))
   }
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -194,11 +194,9 @@ object Input {
     def encode(data: Instant): Chunk[String] = Chunk.single(wrap(data.toEpochMilli.toString))
   }
 
-  case object WeightsInput extends Input[Weights] {
-    def encode(data: Weights): Chunk[String] =
-      data.weights.foldLeft(Chunk.single(wrap("WEIGHTS")): Chunk[String])((acc, a) =>
-        acc ++ Chunk.single(wrap(a.toString))
-      )
+  case object WeightsInput extends Input[Seq[Double]] {
+    def encode(data: Seq[Double]): Chunk[String] =
+      data.foldLeft(Chunk.single(wrap("WEIGHTS")): Chunk[String])((acc, a) => acc ++ Chunk.single(wrap(a.toString)))
   }
 
   final case class Tuple2[-A, -B](_1: Input[A], _2: Input[B]) extends Input[(A, B)] {
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -194,8 +194,8 @@ object Input {
     def encode(data: Instant): Chunk[String] = Chunk.single(wrap(data.toEpochMilli.toString))
   }
 
-  case object WeightsInput extends Input[Seq[Double]] {
-    def encode(data: Seq[Double]): Chunk[String] =
+  case object WeightsInput extends Input[List[Double]] {
+    def encode(data: List[Double]): Chunk[String] =
       data.foldLeft(Chunk.single(wrap("WEIGHTS")): Chunk[String])((acc, a) => acc ++ Chunk.single(wrap(a.toString)))
   }
 
--- a/redis/src/main/scala/zio/redis/Output.scala
+++ b/redis/src/main/scala/zio/redis/Output.scala
@@ -163,13 +163,13 @@ object Output {
 
   case object StringOutput extends Output[String] {
     protected def tryDecode(text: String): String =
-      if (text.startsWith("$"))
+      if (text.startsWith("+"))
         parse(text)
       else
         throw ProtocolError(s"$text isn't a simple string.")
 
     private[this] def parse(text: String): String =
-      text.substring(2, text.length).filter(_ >= ' ')
+      text.substring(1, text.length - 2)
   }
 
   case object MultiStringOutput extends Output[String] {
--- a/redis/src/main/scala/zio/redis/Output.scala
+++ b/redis/src/main/scala/zio/redis/Output.scala
@@ -163,13 +163,13 @@ object Output {
 
   case object StringOutput extends Output[String] {
     protected def tryDecode(text: String): String =
-      if (text.startsWith("+"))
+      if (text.startsWith("$2"))
         parse(text)
       else
         throw ProtocolError(s"$text isn't a simple string.")
 
     private[this] def parse(text: String): String =
-      text.substring(1, text.length - 2)
+      text.substring(2, text.length).filter(_ >= ' ')
   }
 
   case object MultiStringOutput extends Output[String] {
--- a/redis/src/main/scala/zio/redis/Output.scala
+++ b/redis/src/main/scala/zio/redis/Output.scala
@@ -163,7 +163,7 @@ object Output {
 
   case object StringOutput extends Output[String] {
     protected def tryDecode(text: String): String =
-      if (text.startsWith("$2"))
+      if (text.startsWith("$"))
         parse(text)
       else
         throw ProtocolError(s"$text isn't a simple string.")
--- a/redis/src/main/scala/zio/redis/api/SortedSets.scala
+++ b/redis/src/main/scala/zio/redis/api/SortedSets.scala
@@ -30,7 +30,7 @@ trait SortedSets {
         IncrementInput,
         NonEmptyList(MemberScoreInput)
       ),
-      DoubleOutput
+      OptionalOutput(DoubleOutput)
     )
 
   final val zCard   = RedisCommand("ZCARD", StringInput, LongOutput)
--- a/redis/src/main/scala/zio/redis/api/SortedSets.scala
+++ b/redis/src/main/scala/zio/redis/api/SortedSets.scala
@@ -29,9 +29,9 @@ trait SortedSets {
     "ZINTERSTORE",
     Tuple4(
       StringInput,
-      NonEmptyList(StringInput),
-      OptionalInput(NonEmptyList(DoubleInput)),
-      OptionalInput(AggregateInput)
+      NonEmptyListPrecededByCount(StringInput),
+      OptionalInput(AggregateInput),
+      OptionalInput(WeightsInput)
     ),
     LongOutput
   )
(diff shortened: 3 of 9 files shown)


mijicd:
@maciejbak85 Don't worry, I've converted it to draft, change its state once you're ready :).

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -139,6 +139,13 @@ object Input {
       (data._1 :: data._2).foldLeft(Chunk.empty: Chunk[String])((acc, a) => acc ++ input.encode(a))
   }
 
+  final case class NonEmptyListPrecededByCount[-A](input: Input[A]) extends Input[(A, List[A])] {
+    def encode(data: (A, List[A])): Chunk[String] =
+      (data._1 :: data._2).foldLeft(Chunk.single(wrap((1 + data._2.length).toString)): Chunk[String])((acc, a) =>
```
mijicd, line 144: Let's not introduce this one. I get the idea of calculating `numkeys` automatically, but it's not aligned with the rest of operations we expose.

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -163,13 +163,13 @@ object Output {
 
   case object StringOutput extends Output[String] {
     protected def tryDecode(text: String): String =
-      if (text.startsWith("+"))
+      if (text.startsWith("$"))
```
mijicd, line 166: What happened here :confused: ?

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -163,13 +163,13 @@ object Output {
...
         parse(text)
       else
         throw ProtocolError(s"$text isn't a simple string.")
 
     private[this] def parse(text: String): String =
-      text.substring(1, text.length - 2)
+      text.substring(2, text.length).filter(_ >= ' ')
```
mijicd, line 172: Same here.

redis/src/main/scala/zio/redis/api/SortedSets.scala
```diff
@@ -2,24 +2,25 @@ package zio.redis.api
...
-    Tuple5(
-      StringInput,
-      OptionalInput(UpdateInput),
-      OptionalInput(ChangedInput),
-      OptionalInput(IncrementInput),
-      NonEmptyList(MemberScoreInput)
-    ),
-    LongOutput
-  )
+  final def zAdd[T](output: Output[T] = LongOutput) =
```
mijicd, line 12: Please start from `A`. Also, passed like this, any output (even invalid one) will suffice.

redis/src/main/scala/zio/redis/api/SortedSets.scala
```diff
@@ -74,19 +75,19 @@ trait SortedSets {
...
   final val zUnionStore = RedisCommand(
     "ZUNIONSTORE",
     Tuple4(
       StringInput,
-      NonEmptyList(StringInput),
-      OptionalInput(NonEmptyList(DoubleInput)),
-      OptionalInput(AggregateInput)
+      NonEmptyListPrecededByCount(StringInput),
+      OptionalInput(AggregateInput),
+      OptionalInput(WeightsInput)
```
mijicd, line 90: This isn't the default order as per the docs.

redis/src/main/scala/zio/redis/options/SortedSets.scala
```diff
@@ -101,4 +101,7 @@ trait SortedSets {
   }
 
   type WithScores = WithScores.type
+
+  sealed case class Weights(weights: Seq[Long])
```
mijicd, line 105: This should not be needed at all. You can still use `DoubleInput` (not longs) and prepend `WEIGHTS` during command encoding.

mijicd:
@maciejbak85 it looks like you tackled #97 as well, so please mark it in the PR description :pray:

redis/src/main/scala/zio/redis/Output.scala
maciejbak85: it just did not work this way ;) Redis returns $1 or $2 here

redis/src/main/scala/zio/redis/Output.scala
maciejbak85: Sample redis output looks for example:
```
$2
 10
```

redis/src/main/scala/zio/redis/Output.scala
mijicd: The thing is, this is simple string output, it should start with `+`. We have `MultiStringOutput` that already works with `$` afaik.

redis/src/main/scala/zio/redis/api/SortedSets.scala
maciejbak85: hmm will check it

redis/src/main/scala/zio/redis/api/SortedSets.scala
maciejbak85: indeed, hmm I need to think about some restriction only to String or Long then

redis/src/main/scala/zio/redis/Output.scala
maciejbak85: well I did this change based on redis results, what can I say :) otherwise zScore was not working, afaik this StringOutput is not used in other places

redis/src/main/scala/zio/redis/Input.scala
maciejbak85: @mijicd u sure ? as a library user I would expect by lib to make it automatically

redis/src/main/scala/zio/redis/Output.scala
mijicd: Oh, you are right. Since this is the case, I propose to use `MultiStringOutput` here, and the do the "rename & nuke" in the follow-up pull request.

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -187,6 +194,11 @@ object Input {
     def encode(data: Instant): Chunk[String] = Chunk.single(wrap(data.toEpochMilli.toString))
   }
 
+  case object WeightsInput extends Input[Seq[Double]] {
+    def encode(data: Seq[Double]): Chunk[String] =
```
mijicd, line 198: Please no `Seq` :).

redis/src/main/scala/zio/redis/api/SortedSets.scala
maciejbak85: @mijicd do you have maybe some idea, vision about this ?

redis/src/main/scala/zio/redis/Input.scala
maciejbak85: why not ? ;)

redis/src/main/scala/zio/redis/api/SortedSets.scala
maciejbak85: ah well now I look at this another time, and dont like this idea ;)
Because each time user will need to be sure about redis output.

zAddWithIncr would be clear enough, about the output type at least.
One method -> one return type

wdyt ?

redis/src/main/scala/zio/redis/Input.scala
mijicd: I am sure. It is not aligned with the rest of the API. While it is quite neat, I would rather be consistent, and polish this across the API in the follow-up.

redis/src/main/scala/zio/redis/api/SortedSets.scala
mijicd: We shouldn't propagate this to "user land" as it is already determined by presence or absence of increment input. Therefore, let's remove it from here, and determine it during parsing. I think @aleksandarskrbic did something similar in his latest iteration on hashes / geo test.

redis/src/main/scala/zio/redis/Input.scala
mijicd: To be perfectly clear, I think the change you propose is way friendlier, and it would be cool if you can open up an issue about it. We can use that issue to summarize all similar problems and get rid of them together.

redis/src/main/scala/zio/redis/api/SortedSets.scala
mijicd: We have a race condition in this conversation :). I'm fine with that approach, but let's see if @aleksandarskrbic and @anovakovic01 can recall the similar situation we had and the solution for it :).

maciejbak85: ok cool :)

redis/src/main/scala/zio/redis/Input.scala
maciejbak85: ok so for now, user will need to add List size as a param, but we will create new issue for it

redis/src/main/scala/zio/redis/api/SortedSets.scala
aleksandarskrbic: @maciejbak85 I've stumbled on a similar problem while working on Hash API, specifically HMGET command since we want to return a Map[Field, Value]. Problem is that Redis will return just a list of values. We decided to somehow wire together input and output so while parsing the response, we have information about a command and we can extract fields that users passed as arguments. This was an idea, I haven’t started implementation yet.

anovakovic01: I usually had situations where I could place everything under a single output type. In this case, you could create sum type which will represent an integer or double output. So something like this:
```scala
sealed trait ZAddReply
object ZAddReply {
  case class Number(value: Long) extends ZAddReply
  case class Score(value: Double) extends ZAddReply
}
```
and then just create `case object ZAddOutput extends Output[ZAddReply]`. In this case, `zAdd` will have `OptionalOutput(ZAddOutput)` as output parameter. WDYT? @maciejbak85 @mijicd

maciejbak85: @anovakovic01 minus I see is that user needs to be aware of what type redis returns, library user should not bother about it

mijicd: Well, yes and no, we can parse the output in sort of alternative / orElse fashion. Given possible terms, try them out. Still, it might be cumbersome to propagate.

redis/src/main/scala/zio/redis/api/SortedSets.scala
```diff
@@ -8,18 +8,30 @@ trait SortedSets {
...
+    RedisCommand(
+      "ZADD",
+      Tuple5(
+        StringInput,
+        OptionalInput(UpdateInput),
+        OptionalInput(ChangedInput),
+        IncrementInput,
+        NonEmptyList(MemberScoreInput)
+      ),
+      StringOutput
```
anovakovic01, line 33: I think that this one should be `OptionalOutput(DoubleOutput)`.

anovakovic01 (changes requested):

redis/src/main/scala/zio/redis/api/SortedSets.scala
maciejbak85: why ? if you try redis zAdd with INCR, it will return you a string type not double ;)

anovakovic01: Looks like it returns double as bulk string:
```
If the INCR option is specified, the return value will be Bulk string reply:

The new score of member (a double precision floating point number) represented as string, or nil if the operation was aborted (when called with either the XX or the NX option).
```

maciejbak85: @anovakovic01 well try in redis console ;)

btw.
https://redis.io/commands/zadd
> If the INCR option is specified, the return value will be Bulk string reply

maciejbak85: > 127.0.0.1:6379> ZADD set1 1 a 2 b
> (integer) 2
> 127.0.0.1:6379> ZADD set1 INCR 2 a
> "3"

anovakovic01: I've tried it:
```
ZADD myzset INCR 1 "one" 
$1
1
ZADD myzset INCR 4.2 "two" 
$18
4.2000000000000002
```

Also, on the same link, right below the line that you've just sent it says that it is `a double precision floating point number`, it's just represented as string.

maciejbak85: hmm what redis version do you have ?

maciejbak85: ah you are right, I see now implementation of DoubleOutput, wonder why redis-cli is not showimg me those $18 etc. will fix it in a sec

anovakovic01: > wonder why redis-cli is not showimg me those $18 etc. will fix it in a sec

Try sending requests using telnet, then you should get protocol level responses.

redis/src/main/scala/zio/redis/api/SortedSets.scala
```diff
@@ -8,18 +8,30 @@ trait SortedSets {
...
+    RedisCommand(
+      "ZADD",
+      Tuple5(
+        StringInput,
+        OptionalInput(UpdateInput),
+        OptionalInput(ChangedInput),
+        IncrementInput,
+        NonEmptyList(MemberScoreInput)
+      ),
+      DoubleOutput
```
anovakovic01, line 33: Just one more change: add `OptionalOutput(DoubleOutput)` to cover cases when output is `$-1\n\r` (which is equivalent to `nil` in redis-cli).
```
ZADD myzset XX INCR 3.2 "three"
$-1

```

redis/src/main/scala/zio/redis/Input.scala
maciejbak85: @mijicd @anovakovic01 
maybe you will know how to fix this:

```
3 more arguments than can be applied to method apply: (a: String, b: Long)(c: String, cs: String*)(d: Option[zio.redis.Aggregate], e: Option[List[Double]])zio.ZIO[zio.redis.RedisExecutor,zio.redis.RedisError,Long] in class Arg2VarargsArg2
            card   <- zInterStore(dest, 2, (first, List(second)), None, Some(List(2, 3)))
```

this happens now when Ive removed NonEmptyListPrecededByCount

```
RedisCommand(
    "ZINTERSTORE",
    Tuple5(
      StringInput,
      LongInput,
      NonEmptyList(StringInput),
      OptionalInput(AggregateInput),
      OptionalInput(WeightsInput)
    ),
    LongOutput
  )
```
some implicts hell ;) trying to investigate it, but its quite hard

anovakovic01 (approved):
LGTM!

redis/src/main/scala/zio/redis/Input.scala
maciejbak85: I think its a conflict between implicit final class Arg5 and Arg2VarargsArg2

maciejbak85: @mijicd are you fine with removing Arg2VarargsArg2 ? if its not used and causes problems

anovakovic01: @maciejbak85 I think that it's not used anywhere, so feel free to remove it (with this PR every command should be covered by tests).

maciejbak85: ok removed then

maciejbak85:
@anovakovic01 can you check again PR ? ;) Ive pushed 2 commits after your approve

redis/src/main/scala/zio/redis/Output.scala
anovakovic01, line 166: I think that what @mijicd suggested here is keep the current state from `master` and use `MultiStringOutput` where you need Bulk string.

redis/src/main/scala/zio/redis/Output.scala
anovakovic01, line 172: Same here.

anovakovic01 (changes requested):
@maciejbak85 Sorry, I missed a few things, just fix these, and you have approval. :)

redis/src/main/scala/zio/redis/Output.scala
maciejbak85: good catch, fixing it

anovakovic01 (approved):
LGTM!

mijicd (approved):
Nice work! :rocket:

Issue #98, opened 2020-08-03
Fix BITFIELD command


anovakovic01:
The current `bitField` implementation has the following problems:
1. it doesn't allow mixed order of different commands like in the example below (order matters).
2. it cannot parse current output because it is an array of integers and if there is overflow failure, it will have NULL multistring output (`$-1`) as an element of an output array.

**Example:**
```
BITFIELD key INCRBY u4 0 1 OVERFLOW FAIL INCRBY u4 0 9223372036854775807 INCRBY u4 0 1 INCRBY u4 0 1
*4
:10
$-1
:11
:12
```

anovakovic01:
I will work on this one.

PR #101, opened 2020-08-06
Test strings API

### LINKED ISSUE #64: Test strings API
Expand the [ApiSpec](https://github.com/zio/zio-redis/blob/master/src/test/scala/zio/redis/ApiSpec.scala) with strings commands.

anovakovic01:
This PR resolves #64.

**Summary:**
1. Add support for `NOT` for `bitOp` command,
2. Update API with support for `varargs` only params,
3. Add new output type for `mGet`,
4. Fix input for `set` command,
5. Fix input and output for `setBit` command.

**Note:** Command `bitField` should be redefined and tests will be added after it has been fixed (#98).

### DIFF
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -92,6 +92,13 @@ object Input {
     }
   }
 
+  case object DurationTTLInput extends Input[Duration] {
+    def encode(data: Duration): Chunk[String] = {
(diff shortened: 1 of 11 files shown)


anovakovic01:
@mijicd Looks like lint job failed because of connection problems.

mijicd:
@anovakovic01 please rebase, it's not due to connection problem, but an issue in sbt launcher.

PR #105, opened 2020-08-10
Fix BITFIELD command

### LINKED ISSUE #98: Fix BITFIELD command

anovakovic01:
This PR fixes #98.

**Summary:**
1. Fixed input and output for `bitField` command.
2. Added tests for the new `ChunkOptionalLongOutput`.
3. Added tests for `bitField` command.

**Note:** It's not possible to call `bitField` with key as the only parameter and get an empty array using current API, but I think that this is not a big issue as it isn't useful use-case.

### DIFF
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -32,15 +32,16 @@ object Input {
   }
 
   case object BitFieldCommandInput extends Input[BitFieldCommand] {
-    def encode(data: BitFieldCommand): Chunk[String] =
+    def encode(data: BitFieldCommand): Chunk[String] = {
+      import BitFieldCommand._
+
       data match {
-        case BitFieldCommand.BitFieldGet(t, o)     => Chunk(wrap("GET"), wrap(t.stringify), wrap(o.toString))
-        case BitFieldCommand.BitFieldSet(t, o, v)  =>
-          Chunk(wrap("SET"), wrap(t.stringify), wrap(o.toString), wrap(v.toString))
-        case BitFieldCommand.BitFieldIncr(t, o, i) =>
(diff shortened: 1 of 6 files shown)


redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -31,23 +31,16 @@ object Input {
...
+  case object BitFieldCommandInput extends Input[BitFieldCommand] {
+    def encode(data: BitFieldCommand): Chunk[String] =
+      data match {
+        case BitFieldCommand.BitFieldGet(t, o)     => Chunk(wrap("GET"), wrap(t.stringify), wrap(o.toString))
+        case BitFieldCommand.BitFieldSet(t, o, v)  =>
+          Chunk(wrap("SET"), wrap(t.stringify), wrap(o.toString), wrap(v.toString))
+        case BitFieldCommand.BitFieldIncr(t, o, i) =>
+          Chunk(wrap("INCRBY"), wrap(t.stringify), wrap(o.toString), wrap(i.toString))
+        case bfo: BitFieldCommand.BitFieldOverflow => Chunk(wrap("OVERFLOW"), wrap(bfo.stringify))
+      }
```
mijicd, line 43: Let's add `import BitFieldCommand._` to this function, it'll make things slightly more readable.

Issue #136, opened 2020-10-16
Move RESP serialisation into RedisExecutor service


quelgar:
I think some changes to `Interpreter` are needed.

The problem is that the current method of receiving responses isn't reliable, and it can't support pipelining. The `unsafeReceive` method just keeps reading until it only partially fills its buffer, assuming that means it has now read the entire response. There's a few ways this can go wrong, but the easiest way to make it fail when using localhost is to `SET` a key with a 1015 byte value. If you `GET` that with zio-redis, the read will very likely hang because the response happens to fit exactly in the buffer.

I can't see a way to fix this with the current `RedisExecutor` API, which accepts and returns RESP serialized strings. The problem with this is you can't reliably tell where a response ends without actually deserializing.

I propose to fix this by moving RESP serialization/deserialization into the executor. This is generic serialization that doesn't depend on which command is being sent. The RESP serialization can be represented with something like this:

```
sealed trait RespValue

object RespValue {
  final case class SimpleString(value: String) extends RespValue

  final case class Error(value: String) extends RespValue

  final case class Integer(value: Long) extends RespValue

  final case class BulkString(value: Chunk[Byte]) extends RespValue

  final case class Array(values: List[RespValue]) extends  RespValue
  
  case object NullString extends RespValue
}
```

and the executor service could be:

```
trait Service {

      def execute(command: RespValue.Array): IO[RedisError, RespValue]

    }
```

The receive then keeps reading bytes until the deserialization produces a complete response value, which can then be provided to the promise, and then moves on to reading the next response (I'm thinking the deserializaition can be implemented as `ZTransducer`). This will also allow support for pipelining, where the responses will not remotely align with the read buffer.

This will require changes to the `Input` serialization and `Output` deserialization as it exists today, but it should make them a little simpler and separate concerns nicely. For example, `DurationSecondsOutput` can just deal with converting the integer value to a duration, and will no longer need to know the RESP details like the string starting with `:` and parsing the rest to an int.

Thoughts?

mijicd:
This proposal makes sense, but it will require significant changes in the current codebase. It makes sense to prioritize it over transactions and streams to avoid duplicated work. Also, it would be great if we can somehow parallelize it.

mijicd:
Closed by #139

PR #139, opened 2020-10-22
Rework serialisation and interpreter

quelgar:
First attempt at addressing issues discussed in #136.

In master the RESP ([REdis Serialization Protocol](https://redis.io/topics/protocol)) handling is mixed in with the command-specific handling in `Input.scala` and `Output.scala`. This change splits out the RESP serialisation (which works the same way for all commands) into `RespValue.scala`, which constructs an ADT description of the the RESP values.

This produced major changes in `Output.scala`, as the command-specific processing is no longer examining raw RESP strings, but is instead pattern matching on the `RespValue` ADT.

Interpreter has been rewritten to accept and return `RespValue`s. It now uses `AsynchronousSocketChannel`, as that's the simplest way to avoid busy-waiting on a non-blocking socket, without needing blocking threads. More work is likely needed to gracefully handle IO exceptions.

The RESP deserialisation is implemented in the most straight-forward way, processing one byte at a time. This will be slow. At some point it will need to be replaced by Chunk-based sinks to perform decently.

More tests of the RESP serialisation/deserialisation are needed. It's nice that this can now be tested independently
of specific commands. Command tests that were merely testing serialisation have been removed.

A note on `RespValue.BulkString` - this uses `Chunk[Byte]` as its representation, instead of `String`. One of my own major use-cases for Redis is storing raw Protobuf objects, which string doesn't support well. I think a number of the commands should also be changed from `String` to `Chunk[Byte]`, with convenience APIs to accept and return strings. However, this PR is already big enough, so I haven't changed anything in the public API.

### DIFF
--- a/redis/src/main/scala/zio/redis/Interpreter.scala
+++ b/redis/src/main/scala/zio/redis/Interpreter.scala
@@ -16,7 +16,7 @@ trait Interpreter {
 
   object RedisExecutor {
 
-    val defaultPort = 6379
+    private[redis] val DefaultPort = 6379
 
     trait Service {
 
@@ -24,12 +24,7 @@ trait Interpreter {
 
     }
 
-    /**
-     * This is to make better use of pipelining by writing requests to the socket in batches.
-     * Ideally this*(average RESP serialized message size) should equal the buffer size used by `ByteStream`,
-     * but we can only guesstimate.
-     */
-    private val requestQueueSize = 20
+    private val RequestQueueSize = 16
 
     private final class Live(
       reqQueue: Queue[Request],
@@ -42,13 +37,15 @@ trait Interpreter {
           .make[RedisError, RespValue]
           .flatMap(promise => reqQueue.offer(Request(command, promise)) *> promise.await)
 
-      private def sendNext(out: Chunk[Byte] => IO[IOException, Unit]): IO[RedisError, Unit] =
+      private def sendWith(out: Chunk[Byte] => IO[IOException, Unit]): IO[RedisError, Unit] =
         reqQueue.takeBetween(1, Int.MaxValue).flatMap { reqs =>
           val bytes = Chunk.fromIterable(reqs).flatMap(req => RespValue.Array(req.command).serialize)
           out(bytes)
             .mapError(RedisError.IOError)
-            .tapError(e => IO.foreach(reqs)(_.promise.fail(e)))
-            .zipLeft(IO.foreach(reqs)(req => resQueue.offer(req.promise)))
+            .tapBoth(
+              e => IO.foreach_(reqs)(_.promise.fail(e)),
+              _ => IO.foreach_(reqs)(req => resQueue.offer(req.promise))
+            )
         }
 
       private def runReceive(inStream: Stream[IOException, Byte]): IO[RedisError, Unit] =
@@ -66,41 +63,39 @@ trait Interpreter {
         byteStream
           .mapError(RedisError.IOError)
           .use { rwBytes =>
-            sendNext(rwBytes.write).forever race runReceive(rwBytes.read)
+            sendWith(rwBytes.write).forever race runReceive(rwBytes.read)
           }
           .tapError { e =>
-            IO.effectTotal(println(s"Reconnecting due to error: $e")) *>
-              resQueue.takeAll.flatMap(IO.foreach(_)(_.fail(e)))
+            // TODO - log that a reconnection is happening
+            resQueue.takeAll.flatMap(IO.foreach_(_)(_.fail(e)))
           }
           .retryWhile(Function.const(true))
-          .tapCause(c => IO.effectTotal(Console.err.println(s"Executor exiting: $c")))
-
     }
 
-    def live: ZLayer[ByteStream, RedisError.IOError, RedisExecutor] =
+    private def fromBytestream: ZLayer[ByteStream, RedisError.IOError, RedisExecutor] =
       ZLayer.fromServiceManaged { env =>
         for {
-          reqQueue <- Queue.bounded[Request](requestQueueSize).toManaged_
+          reqQueue <- Queue.bounded[Request](RequestQueueSize).toManaged_
           resQueue <- Queue.unbounded[Promise[RedisError, RespValue]].toManaged_
           live      = new Live(reqQueue, resQueue, env.connect)
           _        <- live.run.forkManaged
         } yield live
       }
 
-    def liveServer(address: SocketAddress): Layer[RedisError.IOError, RedisExecutor] =
-      ByteStream.socket(address).mapError(RedisError.IOError) >>> live
+    def live(address: SocketAddress): Layer[RedisError.IOError, RedisExecutor] =
+      ByteStream.socket(address).mapError(RedisError.IOError) >>> fromBytestream
 
-    def liveServer(host: String, port: Int = defaultPort): Layer[RedisError.IOError, RedisExecutor] =
-      ByteStream.socket(host, port).mapError(RedisError.IOError) >>> live
+    def live(host: String, port: Int = DefaultPort): Layer[RedisError.IOError, RedisExecutor] =
+      ByteStream.socket(host, port).mapError(RedisError.IOError) >>> fromBytestream
 
-    def liveLoopback(port: Int = defaultPort): Layer[RedisError.IOError, RedisExecutor] =
-      ByteStream.socketLoopback(port).mapError(RedisError.IOError) >>> live
+    def loopback(port: Int = DefaultPort): Layer[RedisError.IOError, RedisExecutor] =
+      ByteStream.socketLoopback(port).mapError(RedisError.IOError) >>> fromBytestream
 
   }
 
-  type ByteStream = Has[ByteStream.Service]
+  private type ByteStream = Has[ByteStream.Service]
 
-  object ByteStream {
+  private[redis] object ByteStream {
 
     val ResponseBufferSize = 1024
 
@@ -108,7 +103,7 @@ trait Interpreter {
 
     trait Service {
 
-      def connect: Managed[IOException, ReadWriteBytes]
+      val connect: Managed[IOException, ReadWriteBytes]
 
     }
 
@@ -120,24 +115,22 @@ trait Interpreter {
 
     }
 
-    /**
-     * Adapts `ComplectionHandler` async channel operations with interruption support.
-     */
-    private def effectAsyncChannel[C <: Channel, T](
+    private def completionHandlerCallback[A](k: IO[IOException, A] => Unit): CompletionHandler[A, Any] =
+      new CompletionHandler[A, Any] {
+        def completed(result: A, u: Any): Unit = k(IO.succeedNow(result))
+
+        def failed(t: Throwable, u: Any): Unit =
+          t match {
+            case e: IOException => k(IO.fail(e))
+            case _              => k(IO.die(t))
+          }
+      }
+
+    private def effectAsyncChannel[C <: Channel, A](
       channel: C
-    )(op: C => CompletionHandler[T, Any] => Any): IO[IOException, T] =
+    )(op: C => CompletionHandler[A, Any] => Any): IO[IOException, A] =
       IO.effectAsyncInterrupt { k =>
-        val handler = new CompletionHandler[T, Any] {
-          def completed(result: T, u: Any): Unit = k(IO.succeedNow(result))
-
-          def failed(t: Throwable, u: Any): Unit =
-            t match {
-              case e: IOException => k(IO.fail(e))
-              case _              => k(IO.die(t))
-            }
-        }
-
-        op(channel)(handler)
+        op(channel)(completionHandlerCallback(k))
         Left(IO.effect(channel.close()).ignore)
       }
 
@@ -179,9 +172,9 @@ trait Interpreter {
         Managed.fromAutoCloseable(make).refineToOrDie[IOException]
       }
 
-      override def connect: Managed[IOException, ReadWriteBytes] =
+      override val connect: Managed[IOException, ReadWriteBytes] =
         openChannel.map { channel =>
-          val readChunk: IO[Option[IOException], Chunk[Byte]] =
+          val readChunk =
             (for {
               _     <- IO.effectTotal(readBuffer.clear())
               _     <- effectAsyncChannel[AsynchronousByteChannel, Integer](channel)(c => c.read(readBuffer, null, _))
@@ -198,6 +191,7 @@ trait Interpreter {
               case e: IOException  => Some(e)
             }
 
+          @inline
           def writeChunk(chunk: Chunk[Byte]): IO[IOException, Unit] =
             IO.when(chunk.nonEmpty) {
               IO.effectTotal {
@@ -214,9 +208,9 @@ trait Interpreter {
             }
 
           new ReadWriteBytes {
-            override def read = Stream.repeatEffectChunkOption(readChunk)
+            def read: Stream[IOException, Byte] = Stream.repeatEffectChunkOption(readChunk)
 
-            override def write(chunk: Chunk[Byte]) = writeChunk(chunk)
+            def write(chunk: Chunk[Byte]): IO[IOException, Unit] = writeChunk(chunk)
           }
         }
 
--- a/redis/src/main/scala/zio/redis/Interpreter.scala
+++ b/redis/src/main/scala/zio/redis/Interpreter.scala
@@ -1,135 +1,236 @@
 package zio.redis
 
-import java.io.IOException
-import java.net.{ InetSocketAddress, StandardSocketOptions }
+import java.io.{ EOFException, IOException }
+import java.net.{ InetAddress, InetSocketAddress, SocketAddress, StandardSocketOptions }
 import java.nio.ByteBuffer
-import java.nio.channels.SocketChannel
-import java.nio.charset.StandardCharsets.UTF_8
-import java.util.concurrent.atomic.AtomicBoolean
-
-import scala.collection.mutable.ArrayBuilder
+import java.nio.channels.{ AsynchronousByteChannel, AsynchronousSocketChannel, Channel, CompletionHandler }
 
 import zio._
-import zio.blocking._
+import zio.stream.Stream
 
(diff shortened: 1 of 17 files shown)


quelgar:
I meant to click Draft, doh.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
+          .mapError(RedisError.IOError)
+          .use { rwBytes =>
+            sendNext(rwBytes.write).forever race runReceive(rwBytes.read)
+          }
+          .tapError { e =>
+            IO.effectTotal(println(s"Reconnecting due to error: $e")) *>
+              resQueue.takeAll.flatMap(IO.foreach(_)(_.fail(e)))
+          }
+          .retryWhile(Function.const(true))
+          .tapCause(c => IO.effectTotal(Console.err.println(s"Executor exiting: $c")))
```
mijicd, line 76: Same remark as for the print above.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
+
     }
 
-    def live(host: String, port: Int): ZLayer[Blocking, IOException, RedisExecutor] =
+    def live: ZLayer[ByteStream, RedisError.IOError, RedisExecutor] =
```
mijicd, line 80: I wouldn't leak `ByteStream`.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
-    private[this] def connect(host: String, port: Int): Managed[IOException, Connection] = {
-      val makeChannel =
-        IO {
-          val channel = SocketChannel.open(new InetSocketAddress(host, port))
-          channel.configureBlocking(false)
-          channel.setOption(StandardSocketOptions.SO_KEEPALIVE, Boolean.box(true))
-          channel.setOption(StandardSocketOptions.TCP_NODELAY, Boolean.box(true))
-          channel
-        }
+    def liveServer(address: SocketAddress): Layer[RedisError.IOError, RedisExecutor] =
```
mijicd, line 90: I would drop this.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
-          pendingRequests <- Queue.unbounded[Request].toManaged_
-          readinessFlag    = new AtomicBoolean(true)
-          response         = ByteBuffer.allocate(ResponseBufferSize)
-        } yield new Connection(channel, pendingRequests, readinessFlag, response)
+    def liveServer(host: String, port: Int = defaultPort): Layer[RedisError.IOError, RedisExecutor] =
+      ByteStream.socket(host, port).mapError(RedisError.IOError) >>> live
 
-      connection.refineToOrDie[IOException]
-    }
+    def liveLoopback(port: Int = defaultPort): Layer[RedisError.IOError, RedisExecutor] =
```
mijicd, line 96: I'd rename this to `loopback`.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
+      ByteStream.socket(address).mapError(RedisError.IOError) >>> live
 
-      val connection =
-        for {
-          channel         <- Managed.fromAutoCloseable(makeChannel)
-          pendingRequests <- Queue.unbounded[Request].toManaged_
-          readinessFlag    = new AtomicBoolean(true)
-          response         = ByteBuffer.allocate(ResponseBufferSize)
-        } yield new Connection(channel, pendingRequests, readinessFlag, response)
+    def liveServer(host: String, port: Int = defaultPort): Layer[RedisError.IOError, RedisExecutor] =
```
mijicd, line 93: Please rename this to `live` and inline the snippet given above.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
 
-          request.result.unsafeDone(exchange.refineToOrDie[RedisError])
-        }
+  type ByteStream = Has[ByteStream.Service]
 
-      private def unsafeReceive(): String = {
-        val builder      = ArrayBuilder.make[Array[Byte]]
-        var readBytes    = 0
-        var responseSize = 0
+  object ByteStream {
```
mijicd, line 103: I'd make this guy package private.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
+      def write(chunk: Chunk[Byte]): IO[IOException, Unit]
 
-          System.arraycopy(chunk, 0, data, j, chunk.length)
+    }
 
-          i += 1
-          j += chunk.length
+    /**
+     * Adapts `ComplectionHandler` async channel operations with interruption support.
+     */
```
mijicd, line 125: I'd remove the scaladoc.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
+      def execute(command: Chunk[RespValue.BulkString]): IO[RedisError, RespValue]
+
+    }
+
+    /**
+     * This is to make better use of pipelining by writing requests to the socket in batches.
+     * Ideally this*(average RESP serialized message size) should equal the buffer size used by `ByteStream`,
+     * but we can only guesstimate.
+     */
+    private val requestQueueSize = 20
```
mijicd, line 32: Please remove the scaladoc, rename the constant to upper camel case and make it a power of two.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
 
-          System.arraycopy(chunk, 0, data, j, chunk.length)
+    }
 
-          i += 1
-          j += chunk.length
+    /**
+     * Adapts `ComplectionHandler` async channel operations with interruption support.
+     */
+    private def effectAsyncChannel[C <: Channel, T](
```
mijicd, line 126: Please change `T` to `A`.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
+          def failed(t: Throwable, u: Any): Unit =
+            t match {
+              case e: IOException => k(IO.fail(e))
+              case _              => k(IO.die(t))
+            }
         }
 
-        new String(data, UTF_8)
+        op(channel)(handler)
+        Left(IO.effect(channel.close()).ignore)
```
mijicd, line 141: Isn't the channel created in managed?

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
+      IO.effectAsyncInterrupt { k =>
+        val handler = new CompletionHandler[T, Any] {
+          def completed(result: T, u: Any): Unit = k(IO.succeedNow(result))
+
+          def failed(t: Throwable, u: Any): Unit =
+            t match {
+              case e: IOException => k(IO.fail(e))
+              case _              => k(IO.die(t))
+            }
         }
```
mijicd, line 138: This can be factored out.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
+                       channel
+                     }
+          _       <- effectAsyncChannel[AsynchronousSocketChannel, Void](channel)(c => c.connect(address, null, _))
+        } yield channel
+        Managed.fromAutoCloseable(make).refineToOrDie[IOException]
+      }
+
+      override def connect: Managed[IOException, ReadWriteBytes] =
+        openChannel.map { channel =>
+          val readChunk: IO[Option[IOException], Chunk[Byte]] =
```
mijicd, line 184:
```suggestion
          val readChunk =
```

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
+                       channel.setOption(StandardSocketOptions.SO_KEEPALIVE, Boolean.box(true))
+                       channel.setOption(StandardSocketOptions.TCP_NODELAY, Boolean.box(true))
+                       channel
+                     }
+          _       <- effectAsyncChannel[AsynchronousSocketChannel, Void](channel)(c => c.connect(address, null, _))
+        } yield channel
+        Managed.fromAutoCloseable(make).refineToOrDie[IOException]
+      }
+
+      override def connect: Managed[IOException, ReadWriteBytes] =
```
mijicd, line 182: It can be val.

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -4,685 +4,290 @@ import zio.Chunk
...
-  private[this] def decodeError(text: String): RedisError =
-    if (text.startsWith("-ERR"))
-      RedisError.ProtocolError(text.drop(4).trim())
-    else if (text.startsWith("-WRONGTYPE"))
-      RedisError.WrongType(text.drop(10).trim())
-    else
-      null
+  final def map[B](f: A => B): Output[B] =
+    new Output[B] {
+      override protected def tryDecode(respValue: RespValue): B = f(self.tryDecode(respValue))
```
mijicd, line 26:
```suggestion
      protected def tryDecode(respValue: RespValue): B = f(self.tryDecode(respValue))
```

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
+                  .repeatWhileM(_ => IO.effectTotal(writeBuffer.hasRemaining))
+                  .zipRight(writeChunk(remainder))
+              }
+            }
+
+          new ReadWriteBytes {
+            override def read = Stream.repeatEffectChunkOption(readChunk)
+
+            override def write(chunk: Chunk[Byte]) = writeChunk(chunk)
+          }
```
mijicd, line 220:
```suggestion
          new ReadWriteBytes {
            def read: Stream[IOException, Byte] = Stream.repeatEffectChunkOption(readChunk)

            def write(chunk: Chunk[Byte]): IO[IOException, Unit] = writeChunk(chunk)
          }
```

Looking at it again, I would inline `readChunk` and `writeChunk`.

redis/src/main/scala/zio/redis/RespValue.scala
```diff
@@ -0,0 +1,197 @@
...
+          (acc, elem) => acc ++ elem.serialize
+        ))
+      case NullValue         =>
+        nullString
+    }
+  }
+
+}
+
+object RespValue {
```
mijicd, line 35: Rename constants in upper camel case fashion.

redis/src/main/scala/zio/redis/RespValue.scala
```diff
@@ -0,0 +1,197 @@
...
+import zio.{ Chunk, IO, Ref }
+
+sealed trait RespValue extends Any {
+
+  import RespValue._
+
+  final def serialize: Chunk[Byte] = {
+    def simpleString(s: String) = Chunk.fromArray(s.getBytes(StandardCharsets.US_ASCII)) ++ crLf
+
+    this match {
```
mijicd, line 15: I would use self just for consistency.

redis/src/main/scala/zio/redis/RespValue.scala
```diff
@@ -0,0 +1,197 @@
+package zio.redis
+
+import java.nio.charset.StandardCharsets
+
+import zio.stream.{ Sink, ZSink }
+import zio.{ Chunk, IO, Ref }
+
+sealed trait RespValue extends Any {
```
mijicd, line 8: You should be able to remove the extends part.

redis/src/main/scala/zio/redis/RespValue.scala
```diff
@@ -0,0 +1,197 @@
...
+  def array(values: RespValue*): Array = Array(Chunk.fromIterable(values))
+
+  def bulkString(s: String): BulkString = BulkString(Chunk.fromArray(s.getBytes(StandardCharsets.UTF_8)))
+
+  def decodeString(bytes: Chunk[Byte]): String = new String(bytes.toArray, StandardCharsets.UTF_8)
+
+  sealed trait State {
+    import State._
+    final def continue: Boolean =
+      this match {
```
mijicd, line 78: Let's use `self` for consistency.

redis/src/main/scala/zio/redis/RespValue.scala
```diff
@@ -0,0 +1,197 @@
...
+
+  def array(values: RespValue*): Array = Array(Chunk.fromIterable(values))
+
+  def bulkString(s: String): BulkString = BulkString(Chunk.fromArray(s.getBytes(StandardCharsets.UTF_8)))
+
+  def decodeString(bytes: Chunk[Byte]): String = new String(bytes.toArray, StandardCharsets.UTF_8)
+
+  sealed trait State {
+    import State._
+    final def continue: Boolean =
```
mijicd, line 77: I'd rename it to `finished` and update the values accordingly.

redis/src/main/scala/zio/redis/RespValue.scala
```diff
@@ -0,0 +1,197 @@
...
+
+    final case class CrSeen(chunk: Chunk[Char]) extends State
+
+    final case class Complete(chunk: Chunk[Char]) extends State
+
+    case object Failed extends State
+
+  }
+
+  def simpleStringDeserialize: Sink[RedisError.ProtocolError, Byte, Byte, String] = {
```
mijicd, line 96: All sinks here can be vals.

mijicd (changes requested):
I like this, amazing work! Please fix the remarks and it's good to go.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,236 @@
...
+       * Only exits by interruption or defect.
+       */
+      def run: IO[RedisError, Unit] =
+        byteStream
+          .mapError(RedisError.IOError)
+          .use { rwBytes =>
+            sendNext(rwBytes.write).forever race runReceive(rwBytes.read)
+          }
+          .tapError { e =>
+            IO.effectTotal(println(s"Reconnecting due to error: $e")) *>
```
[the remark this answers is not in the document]
quelgar: Yeah, this was a a temporary hack. I think we do want to give some visibility on stuff like this, but it makes sense to tackle it properly later on. I'll put a TODO for now.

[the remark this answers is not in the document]
mijicd: Let's open up an issue instead, once this is merged. TODOs tend to be ignored :).

redis/src/main/scala/zio/redis/Interpreter.scala
quelgar: It would be useful to expose it if we want the option to use something other than a TCP socket for the transport. But I admit I can't think of any situation where that's useful…

redis/src/main/scala/zio/redis/Interpreter.scala
quelgar: I think it's useful, sometimes it's good to have control over the address resolution. Also, it is technically _possible_ for a Java client to connect to Redis over a Unix domain socket (with an extra library), which would require the `SocketAddress` API to work.

mijicd: Hm, ok. Let's keep it, but please rename to `live`.

redis/src/main/scala/zio/redis/Interpreter.scala
quelgar: What's the reason for removing Scaladoc?

mijicd: No need for scaladoc on private stuff imho. However, if you want, let's keep the remark as regular multi-line.

redis/src/main/scala/zio/redis/Interpreter.scala
quelgar: This is for interruption. The only way to interrupt an in-progress I/O operation on a channel is to close the channel (well, the `java.util.concurrent.Future` version _technically_ lets you keep the channel open in a weird state, but it's cleaner just to close it).

redis/src/main/scala/zio/redis/Interpreter.scala
quelgar: I'm not sure what you mean here.

mijicd: Just to pull out the handler creation into separate function. You can keep it here, it's a nitpick.

redis/src/main/scala/zio/redis/RespValue.scala
quelgar: This lets the case classes extend `AnyVal`. I'm not 100% sure how practically useful that is, but it might allow them to be put into chunks without boxing.

redis/src/main/scala/zio/redis/Interpreter.scala
quelgar: Made it private 👍

quelgar:
Ok, I think that's everything.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -1,135 +1,230 @@
...
+                         val array = Array.ofDim[Byte](count)
+                         readBuffer.get(array)
+                         Chunk.fromArray(array)
+                       }
+            } yield chunk).mapError {
+              case _: EOFException => None
+              case e: IOException  => Some(e)
+            }
+
+          @inline
```
mijicd, line 194: Hehe, I meant to inline the function completely, but annotation can work as well.

mijicd (approved):

redis/src/main/scala/zio/redis/Interpreter.scala
quelgar: Oh I see, but it's recursive, so we can't.

PR #141, opened 2020-11-01
Add logging capability

### LINKED ISSUE #140: Add logging capability
I don't think we need to go overboard with logging, but it would be useful to log when connections and reconnections are occurring, and why. Once we have Sentinel support we'd also want to log the outcomes of Sentinel discovery.

It might also help users debugging apps if there was an option to log a trace of commands and responses?

Probably [zio-logging](https://github.com/zio/zio-logging) is the way to go.

anovakovic01:
This PR resolves #140.

**Summary:**
1. Add logs for connecting, reconnecting, and disconnecting operations,
2. Add debug logs for sending commands and receiving results.

### DIFF
--- a/redis/src/main/scala/zio/redis/Interpreter.scala
+++ b/redis/src/main/scala/zio/redis/Interpreter.scala
@@ -4,9 +4,10 @@ import java.io.{ EOFException, IOException }
 import java.net.{ InetAddress, InetSocketAddress, SocketAddress, StandardSocketOptions }
 import java.nio.ByteBuffer
 import java.nio.channels.{ AsynchronousByteChannel, AsynchronousSocketChannel, Channel, CompletionHandler }
+import java.util.UUID
 
 import zio._
-import zio.logging.{ Logger, Logging }
+import zio.logging._
 import zio.stream.Stream
 
 trait Interpreter {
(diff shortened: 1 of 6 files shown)


redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -29,13 +31,20 @@ trait Interpreter {
...
-      byteStream: Managed[IOException, ByteStream.ReadWriteBytes]
+      byteStream: Managed[IOException, ByteStream.ReadWriteBytes],
+      logger: Logger[String]
     ) extends Service {
 
       override def execute(command: Chunk[RespValue.BulkString]): IO[RedisError, RespValue] =
-        Promise
-          .make[RedisError, RespValue]
-          .flatMap(promise => reqQueue.offer(Request(command, promise)) *> promise.await)
+        logger.locally(LogAnnotation.CorrelationId(Some(UUID.randomUUID()))) {
```
mijicd, line 39: I wouldn't do any logging here.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -66,30 +75,34 @@ trait Interpreter {
             sendWith(rwBytes.write).forever race runReceive(rwBytes.read)
           }
           .tapError { e =>
-            // TODO - log that a reconnection is happening
-            resQueue.takeAll.flatMap(IO.foreach_(_)(_.fail(e)))
+            logger.warn(s"Reconnecting due to error: $e") *>
+              resQueue.takeAll.flatMap(IO.foreach_(_)(_.fail(e)))
           }
           .retryWhile(Function.const(true))
+          .tapError { e =>
```
mijicd, line 82: You can drop the block syntax here.

mijicd (commented):
Make sure to reformat the sources.

redis/src/main/scala/zio/redis/Interpreter.scala
anovakovic01: Removed.

redis/src/main/scala/zio/redis/Interpreter.scala
anovakovic01: Done.

mijicd (approved):

Issue #169, opened 2020-11-20
Develop testable RedisExecutor


mijicd:
Design and implement a test Redis executor (check out [this module](https://github.com/zio/zio-redis/blob/master/redis/src/main/scala/zio/redis/Interpreter.scala)).

Ideally, the test executor would execute all of the commands backed by some in-memory structures. I propose to expose the test implementation in the same module via additional layer constructors.

Since this ticket will be involving, I propose to come up with a design proposal and then split the implementation work into API-specific pieces.

quelgar:
Just a minor point while I think of it, the `RespValue.NullValue` will need to be split into two, as RESP actually has two kinds of null: a null bulk string and a null array. These are currently both represented by the same value because a client only ever reads null arrays, it never needs to create them.

mschuwalow:
I'll take this one

mijicd:
The design is provided in #180, closing this in favor of:

- https://github.com/zio/zio-redis/issues/190
- https://github.com/zio/zio-redis/issues/191
- https://github.com/zio/zio-redis/issues/192
- https://github.com/zio/zio-redis/issues/193
- https://github.com/zio/zio-redis/issues/194
- https://github.com/zio/zio-redis/issues/195
- https://github.com/zio/zio-redis/issues/196
- https://github.com/zio/zio-redis/issues/197
- https://github.com/zio/zio-redis/issues/198

mschuwalow:
An initial skeleton has been added. Further work is tracked in the following issues:
- [ ] Connection API - https://github.com/zio/zio-redis/issues/190
- [ ] Geo API - https://github.com/zio/zio-redis/issues/191
- [ ] Hashes API - https://github.com/zio/zio-redis/issues/192
- [ ] HyperLogLog API - https://github.com/zio/zio-redis/issues/193
- [ ] Keys API - https://github.com/zio/zio-redis/issues/194
- [ ] Lists API - https://github.com/zio/zio-redis/issues/195
- [ ] Sets API - https://github.com/zio/zio-redis/issues/196
- [ ] SortedSets API - https://github.com/zio/zio-redis/issues/197
- [ ] Strings API - https://github.com/zio/zio-redis/issues/198

PR #180, opened 2020-11-21
Add test executor

mschuwalow:
part of #169 

As we implement more parts of the API we can enable the tests one by one.

Also fixes a bug in the PING command

### DIFF
--- a/redis/src/main/scala/zio/redis/Interpreter.scala
+++ b/redis/src/main/scala/zio/redis/Interpreter.scala
@@ -6,8 +6,10 @@ import java.nio.ByteBuffer
 import java.nio.channels.{ AsynchronousByteChannel, AsynchronousSocketChannel, Channel, CompletionHandler }
 
 import zio._
+import zio.stm._
 import zio.logging._
 import zio.stream.Stream
+import zio.redis.RedisError.ProtocolError
 
 trait Interpreter {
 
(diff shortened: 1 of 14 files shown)


redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -75,6 +77,27 @@ trait Interpreter {
           .tapError(e => logger.error(s"Executor exiting: $e"))
     }
 
+    private final class InMemory(
```
mijicd, line 80: I'd remove `()`.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -75,6 +77,27 @@ trait Interpreter {
...
+    ) extends Service {
+      override def execute(command: Chunk[RespValue.BulkString]): zio.IO[RedisError, RespValue] =
+        for {
+          name   <- ZIO.fromOption(command.headOption).mapError(_ => ProtocolError("Malformed command."))
```
mijicd, line 84:
```suggestion
          name   <- ZIO.fromOption(command.headOption).orElseFail(ProtocolError("Malformed command."))
```

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -86,6 +109,9 @@ trait Interpreter {
           } yield live
       }
 
+    val inMemory: ZLayer[Any, Nothing, RedisExecutor] =
```
mijicd, line 112: I'd rename this to `test`.

redis/src/main/scala/zio/redis/Interpreter.scala
mschuwalow: Makes sense for now. I was thinking that it will have STM data structures in it

mijicd: Ah yes, true. Well, you'll introduce it anyhow, so this is a no-op.

mijicd (approved):

Issue #160, opened 2020-11-20
Implement pub/sub API


mijicd:
Extracted from [this comment](https://github.com/zio/zio-redis/issues/68#issuecomment-691777714) authored by @regis-leray.

### Commands

- [ ] PSUBSCRIBE pattern [pattern ...] Listen for messages published to channels matching the given patterns
- [ ] PUBSUB subcommand [argument [argument ...]] Inspect the state of the Pub/Sub subsystem
- [ ] PUBLISH channel message Post a message to a channel
- [ ] PUNSUBSCRIBE [pattern [pattern ...]] Stop listening for messages posted to channels matching the given patterns
- [ ] SUBSCRIBE channel [channel ...] Listen for messages published to the given channels
- [ ] UNSUBSCRIBE [channel [channel ...]] Stop listening for messages posted to the given channels

### Tips

- Define the API in `zio.redis.api`.
- Command-specific options should be defined in `zio.redis.options`.
- Additional inputs and outputs can be defined in `zio.redis.Input` and `zio.redis.Output`, respectively.
- Redis API evolves, therefore the commands listed above might be slightly changed. Double check [the official docs](https://redis.io/commands) before starting with implementation. 
- Use telnet to understand command protocol details.
- Write inputs and outputs tests in `zio.redis.InputSpec` and `zio.redis.OutputSpec`, respectively.
- Write integration tests by expanding the `zio.redis.ApiSpec`.
- If you feel that the pull request size is growing out of control, feel free to split it but make sure to link this issue in each of the related PRs.

visox:
Hi, i can go for it. I probably go from top to bottom but its not that much so if i manage to do one i probably i can make it all.

mijicd:
@visox That sounds good. And I agree, it's perfectly fine to split this one, as you might need to adjust the executor occasionally to support it properly (e.g. separate queue / connection).

visox:
i made WIP pr with some question, plz take a look https://github.com/zio/zio-redis/pull/181 @mijicd 
thx, will continue tomorrow.

KadekM:
@mijicd @anovakovic01 

I'm thinking of tackling this, could you check some rough ideas?

> Once the client enters the subscribed state it is not supposed to issue any other commands, except for additional SUBSCRIBE, PSUBSCRIBE, UNSUBSCRIBE, PUNSUBSCRIBE, PING, RESET and QUIT commands.

[...]

PR #181, opened 2020-11-21
WIP issue/160 [hackathon] add pubsub commands

visox:
WIP as i have questions related to my solution, also not complete yet (not tested and missing tests).

Problems/questions
* The command PUBSUB has a sub command, which makes things harder because the input/output type depends on the type of the subcommand. I tried creating a sealed trait for the sub commands but i was unable to create a nice solution with that. Currently i made simply 3 functions (`pubSubChannels`, `pubSubNumSub ` and `pubSubNumPat`).
* What is the right approach to reuse an Output ? See `KeyLongValueOutput` uses `KeyValueOutput`

Thx

### DIFF
--- a/redis/src/main/scala/zio/redis/Output.scala
+++ b/redis/src/main/scala/zio/redis/Output.scala
@@ -3,6 +3,8 @@ package zio.redis
 import zio.Chunk
 import zio.duration._
 
+import scala.util.Try
+
 sealed trait Output[+A] {
 
   self =>
@@ -290,4 +292,20 @@ object Output {
       }
   }
 
+  case object KeyLongValueOutput extends Output[Map[String, Long]] {
+    protected def tryDecode(respValue: RespValue): Map[String, Long] = {
+      val mapOutput = KeyValueOutput.unsafeDecode(respValue)
+
+      val res = mapOutput
+        .mapValues(value => Try(value.toLong).toOption)
+        .collect {
+          case (key, Some(value)) => key -> value
+        }
+
+      if (res.size != mapOutput.size)
+        throw ProtocolError(s"$mapOutput doesn't have a number after a string")
+      else
+        res
+    }
+  }
 }
--- a/redis/src/main/scala/zio/redis/api/PubSub.scala
+++ b/redis/src/main/scala/zio/redis/api/PubSub.scala
@@ -1,0 +1,41 @@
+package zio.redis.api
+
+import zio.redis.Input._
+import zio.redis.Output._
+import zio.redis._
+import zio.{ Chunk, ZIO }
+
+trait PubSub {
+  import PubSub._
+
+  final def pSubscribe(a: String, as: String*): ZIO[RedisExecutor, RedisError, Unit] = PSubscribe.run((a, as.toList))
+
+  final def pubSubChannels(a: String): ZIO[RedisExecutor, RedisError, Chunk[String]] =
+    PubSubChannels.run(("CHANNELS", List(a)))
+
+  final def pubSubNumSub(a: String, as: String*): ZIO[RedisExecutor, RedisError, Map[String, Long]] =
+    PubSubNumSub.run(("NUMSUB", a :: as.toList))
+
+  final def pubSubNumPat(): ZIO[RedisExecutor, RedisError, Long] = PubSubNumPat.run("NUMPAT")
+
+  final def publish(a: String, b: String): ZIO[RedisExecutor, RedisError, Long] = Publish.run((a, b))
+
+  final def pUnSubscribe(as: String*): ZIO[RedisExecutor, RedisError, Unit] = PUnSubscribe.run(as.toList)
+
+  final def subscribe(a: String, as: String*): ZIO[RedisExecutor, RedisError, Unit] = Subscribe.run((a, as.toList))
+
+  final def unSubscribe(as: String*): ZIO[RedisExecutor, RedisError, Unit] = UnSubscribe.run(as.toList)
+
+}
+
+private object PubSub {
+  final val PSubscribe     = RedisCommand("PSUBSCRIBE", NonEmptyList(StringInput), UnitOutput)
+  final val PubSubChannels = RedisCommand("PUBSUB", NonEmptyList(StringInput), ChunkOutput)
+  final val PubSubNumSub   = RedisCommand("PUBSUB", NonEmptyList(StringInput), KeyLongValueOutput)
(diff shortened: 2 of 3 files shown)


redis/src/main/scala/zio/redis/api/PubSub.scala
```diff
@@ -0,0 +1,41 @@
...
+  final def pubSubNumSub(a: String, as: String*): ZIO[RedisExecutor, RedisError, Map[String, Long]] =
+    PubSubNumSub.run(("NUMSUB", a :: as.toList))
+
+  final def pubSubNumPat(): ZIO[RedisExecutor, RedisError, Long] = PubSubNumPat.run("NUMPAT")
+
+  final def publish(a: String, b: String): ZIO[RedisExecutor, RedisError, Long] = Publish.run((a, b))
+
+  final def pUnSubscribe(as: String*): ZIO[RedisExecutor, RedisError, Unit] = PUnSubscribe.run(as.toList)
+
+  final def subscribe(a: String, as: String*): ZIO[RedisExecutor, RedisError, Unit] = Subscribe.run((a, as.toList))
```
anovakovic01, line 25: The problem with `subscribe` is that it keeps an open connection while receiving messages. This could be represented with `ZStream` instead of `ZIO`. I recommend that you try using it through telnet, and then evaluate what changes will be required in order to implement this. You can write your conclusion in issue #160.

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -290,4 +292,20 @@ object Output {
       }
   }
 
+  case object KeyLongValueOutput extends Output[Map[String, Long]] {
+    protected def tryDecode(respValue: RespValue): Map[String, Long] = {
+      val mapOutput = KeyValueOutput.unsafeDecode(respValue)
```
anovakovic01, line 297: Good direction, but you cannot reuse `KeyValueOutput` here, because `KeyValueOutput` expects `RespValue.BulkString` for the value format, while your type should expect `RespValue.Integer`. `RespValue`s are protocol-specific values, so you cannot first parse something as `RespValue.BulkString` (in scala terms `String`), and then cast it to `RespValue.Integer`. You'll have to implement `tryDecode` directly (same logic as `KeyValueOutput`, but instead of `RespValue.BulkString`, for the second value, you should use `RespValue.Integer`).

redis/src/main/scala/zio/redis/api/PubSub.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import zio.redis.Input._
+import zio.redis.Output._
+import zio.redis._
+import zio.{ Chunk, ZIO }
+
+trait PubSub {
+  import PubSub._
+
+  final def pSubscribe(a: String, as: String*): ZIO[RedisExecutor, RedisError, Unit] = PSubscribe.run((a, as.toList))
```
anovakovic01, line 11: Please give meaningful names to the parameters.

redis/src/main/scala/zio/redis/api/PubSub.scala
```diff
@@ -0,0 +1,41 @@
...
+    PubSubChannels.run(("CHANNELS", List(a)))
+
+  final def pubSubNumSub(a: String, as: String*): ZIO[RedisExecutor, RedisError, Map[String, Long]] =
+    PubSubNumSub.run(("NUMSUB", a :: as.toList))
+
+  final def pubSubNumPat(): ZIO[RedisExecutor, RedisError, Long] = PubSubNumPat.run("NUMPAT")
+
+  final def publish(a: String, b: String): ZIO[RedisExecutor, RedisError, Long] = Publish.run((a, b))
+
+  final def pUnSubscribe(as: String*): ZIO[RedisExecutor, RedisError, Unit] = PUnSubscribe.run(as.toList)
```
anovakovic01, line 23: I would name this method `pUnsubscribe`.

redis/src/main/scala/zio/redis/api/PubSub.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  final def subscribe(a: String, as: String*): ZIO[RedisExecutor, RedisError, Unit] = Subscribe.run((a, as.toList))
+
+  final def unSubscribe(as: String*): ZIO[RedisExecutor, RedisError, Unit] = UnSubscribe.run(as.toList)
```
anovakovic01, line 27: I would name this method `unsubscribe`.

redis/src/main/scala/zio/redis/api/PubSub.scala
```diff
@@ -0,0 +1,41 @@
...
+
+}
+
+private object PubSub {
+  final val PSubscribe     = RedisCommand("PSUBSCRIBE", NonEmptyList(StringInput), UnitOutput)
+  final val PubSubChannels = RedisCommand("PUBSUB", NonEmptyList(StringInput), ChunkOutput)
+  final val PubSubNumSub   = RedisCommand("PUBSUB", NonEmptyList(StringInput), KeyLongValueOutput)
+  final val PubSubNumPat   = RedisCommand("PUBSUB", StringInput, LongOutput)
+  final val Publish        = RedisCommand("PUBLISH", Tuple2(StringInput, StringInput), LongOutput)
+  final val PUnSubscribe   = RedisCommand("PUNSUBSCRIBE", ListInput(StringInput), UnitOutput)
```
anovakovic01, line 37: I would name it `PUnsubscribe`.

redis/src/main/scala/zio/redis/api/PubSub.scala
```diff
@@ -0,0 +1,41 @@
...
+  final val Subscribe      = RedisCommand("SUBSCRIBE", NonEmptyList(StringInput), UnitOutput)
+  final val UnSubscribe    = RedisCommand("UNSUBSCRIBE", ListInput(StringInput), UnitOutput)
```
anovakovic01, line 39: I would name it `Unsubscribe`.

anovakovic01 (changes requested):

redis/src/main/scala/zio/redis/api/PubSub.scala
visox: kk

redis/src/main/scala/zio/redis/Output.scala
visox: i see i will write it whole then

mijicd:
Closed due to inactivity.

PR #182, opened 2020-11-21
Get the dump/restore commands working.

### LINKED ISSUE #78: Fix DUMP command
Dump response can't be parsed. Find the root cause and "unignore" the existing test.

quelgar:
The executor now operates with `Chunk[Byte]` instead of `String`, so we can just switch to bytes in the API to pass along the exact bytes that Redis gives us.

Fixes #78


anovakovic01 (approved):
Nice work, thanks @quelgar!

mijicd (approved):

PR #183, opened 2020-11-21
Document strings api #179

### LINKED ISSUE #179: Document strings API
The current strings API (see `zio.redis.api.Strings`) exposes functions with argument names that are generic and meaningless. I propose to rename those to match those used in the [official documentation](https://redis.io/commands). Additionally, it would be great to provide a small, single-sentence scaladoc for them.

justinhj:
Resolves #179

### DIFF
--- a/redis/src/main/scala/zio/redis/api/Strings.scala
+++ b/redis/src/main/scala/zio/redis/api/Strings.scala
@@ -10,68 +10,116 @@ import zio.{ Chunk, ZIO }
 trait Strings {
   import Strings._
 
-  final def append(a: String, b: String): ZIO[RedisExecutor, RedisError, Long] = Append.run((a, b))
+  // Append a value to a key
+  final def append(key: String, value: String): ZIO[RedisExecutor, RedisError, Long] = Append.run((key, value))
 
-  final def bitCount(a: String, b: Option[Range] = None): ZIO[RedisExecutor, RedisError, Long] = BitCount.run((a, b))
+  // Count set bits in a string
+  final def bitCount(key: String, range: Option[Range] = None): ZIO[RedisExecutor, RedisError, Long] =
+    BitCount.run((key, range))
 
+  // Perform arbitrary bitfield integer operations on strings
   final def bitField(
-    a: String,
-    b: BitFieldCommand,
-    bs: BitFieldCommand*
-  ): ZIO[RedisExecutor, RedisError, Chunk[Option[Long]]] = BitField.run((a, (b, bs.toList)))
+    key: String,
+    firstCommand: BitFieldCommand,
+    restCommands: BitFieldCommand*
+  ): ZIO[RedisExecutor, RedisError, Chunk[Option[Long]]] = BitField.run((key, (firstCommand, restCommands.toList)))
 
-  final def bitOp(a: BitOperation, b: String, c: String, cs: String*): ZIO[RedisExecutor, RedisError, Long] =
-    BitOp.run((a, b, (c, cs.toList)))
+  // Perform bitwise operations between strings
+  final def bitOp(
+    operation: BitOperation,
+    destKey: String,
+    firstSrcKey: String,
+    restSrcKeys: String*
+  ): ZIO[RedisExecutor, RedisError, Long] =
+    BitOp.run((operation, destKey, (firstSrcKey, restSrcKeys.toList)))
 
-  final def bitPos(a: String, b: Boolean, c: Option[BitPosRange] = None): ZIO[RedisExecutor, RedisError, Long] =
-    BitPos.run((a, b, c))
+  // Find first bit set or clear in a string
+  final def bitPos(key: String, bit: Boolean, range: Option[BitPosRange] = None): ZIO[RedisExecutor, RedisError, Long] =
+    BitPos.run((key, bit, range))
 
-  final def decr(a: String): ZIO[RedisExecutor, RedisError, Long] = Decr.run(a)
+  // Decrement the integer value of a key by one
+  final def decr(key: String): ZIO[RedisExecutor, RedisError, Long] = Decr.run(key)
 
-  final def decrBy(a: String, b: Long): ZIO[RedisExecutor, RedisError, Long] = DecrBy.run((a, b))
+  // Decrement the integer value of a key by the given number
+  final def decrBy(key: String, decrement: Long): ZIO[RedisExecutor, RedisError, Long] = DecrBy.run((key, decrement))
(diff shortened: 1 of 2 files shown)


redis/src/main/scala/zio/redis/api/Strings.scala
```diff
@@ -10,68 +10,116 @@ import zio.{ Chunk, ZIO }
...
+    MSet.run((firstKeyValue, restKeyValues.toList))
+
+  // Set multiple keys to multiple values only if none of the keys exist
+  final def mSetNx(
+    firstKeyValue: (String, String),
+    restKeyValues: (String, String)*
+  ): ZIO[RedisExecutor, RedisError, Boolean] =
+    MSetNx.run((firstKeyValue, restKeyValues.toList))
+
+  // Set the value and expiration in milliseconds of a key
```
anovakovic01, line 87: In order to have a valid scaladoc comment, you should use `/** <comment here> */` format as you've already used below. Also, try using `@param` to describe method parameters and `@return` to describe a return value. I suggest that you don't use `@return` when the returned ZIO effect has a value of type `Unit`.

Here is one example:
```suggestion
 /** Sets a value and expiration in milliseconds of the specified key.
  *
  *  @param key Redis key identifier
  *  @param milliseconds expire time that is expressed in milliseconds
  *  @param value key value to be set
  */
```

For more details you can read the [documentation](https://docs.scala-lang.org/style/scaladoc.html).

justinhj: Sure I will update it shortly.

redis/src/main/scala/zio/redis/api/Strings.scala
```diff
@@ -10,68 +10,235 @@ import zio.{ Chunk, ZIO }
...
-    MSet.run((a, as.toList))
-
-  final def mSetNx(a: (String, String), as: (String, String)*): ZIO[RedisExecutor, RedisError, Boolean] =
-    MSetNx.run((a, as.toList))
-
-  final def pSetEx(a: String, b: Duration, c: String): ZIO[RedisExecutor, RedisError, Unit] = PSetEx.run((a, b, c))
-
+    key: String,
+    firstCommand: BitFieldCommand,
+    restCommands: BitFieldCommand*
```
anovakovic01, line 41: I'd rename to `command` and `commands`.

redis/src/main/scala/zio/redis/api/Strings.scala
```diff
@@ -10,68 +10,235 @@ import zio.{ Chunk, ZIO }
...
+   * @param destKey Key of destination string to store the result
+   * @param firstSrcKey First source key to apply the operation to
+   * @param restSrcKeys Subsequent source keys to apply the operation to
+   * @return Returns size of the string stored in the destination key, that is equal to the size of the longest input string
+   */
+  final def bitOp(
+    operation: BitOperation,
+    destKey: String,
+    firstSrcKey: String,
+    restSrcKeys: String*
```
anovakovic01, line 56: I'd rename to `srcKey` and `srcKeys`.

redis/src/main/scala/zio/redis/api/Strings.scala
```diff
@@ -10,68 +10,235 @@ import zio.{ Chunk, ZIO }
...
+  final def incrByFloat(key: String, increment: Double): ZIO[RedisExecutor, RedisError, String] =
+    IncrByFloat.run((key, increment))
+
+  /** Get all the values of the given keys
+   *
+    * @param firstKey First key to get
+   * @param restKeys Subsequent keys to get
+   * @return Returns the values of the given keys
+   */
+  final def mGet(firstKey: String, restKeys: String*): ZIO[RedisExecutor, RedisError, Chunk[Option[String]]] =
```
anovakovic01, line 148: I'd rename to `key` and `keys`.

redis/src/main/scala/zio/redis/api/Strings.scala
```diff
@@ -10,68 +10,235 @@ import zio.{ Chunk, ZIO }
...
+    MGet.run((firstKey, restKeys.toList))
+
+  /** Set multiple keys to multiple values
+   *
+    * @param firstKeyValue Tuple of key and value, first one to set
+   * @param restKeyValues Subsequent tuples of key values
+   */
+  final def mSet(
+    firstKeyValue: (String, String),
+    restKeyValues: (String, String)*
```
anovakovic01, line 158: I'd rename to `keyValue` and `keyValues`.

redis/src/main/scala/zio/redis/api/Strings.scala
```diff
@@ -10,68 +10,235 @@ import zio.{ Chunk, ZIO }
...
+
+  /** Set multiple keys to multiple values only if none of the keys exist
+   *
+    * @param firstKeyValue First key value to set
+   * @param restKeyValues Subsequent key values to set
+   * @return 1 if the all the keys were set. 0 if no key was set (at least one key already existed)
+   */
+  final def mSetNx(
+    firstKeyValue: (String, String),
+    restKeyValues: (String, String)*
```
anovakovic01, line 170: I'd rename to `keyValue` and `keyValues`.

anovakovic01 (changes requested):
Good job @justinhj, almost done.

redis/src/main/scala/zio/redis/api/Strings.scala
aleksandarskrbic: @anovakovic01 I like your naming suggestion more, but this style already passed some earlier [PR](https://github.com/zio/zio-redis/pull/185/files). @mijicd We should choose one style and stick to it?

anovakovic01: Good catch. I agree that we should stick to one style only. @mijicd should decide which one.

aleksandarskrbic: I've followed the style of that merged PR, but I could change my PR.

mijicd: Let's go with the proposal from @anovakovic01

justinhj: I will update the names to remove the first and rest prefixes. I will need to open a new PR to do the same in the Sets API

anovakovic01 (approved):

anovakovic01:
@justinhj Nice work. We agreed to modify Scaladoc comment format so that every comment has the first line empty.

PR #200, opened 2020-11-21
Implement streams API

### LINKED ISSUE #157: Implement streams API

### Commands

- [x] XINFO [CONSUMERS key groupname] [GROUPS key] [STREAM key] [HELP] Get information on streams and consumer groups
- [x] XADD key ID field value [field value ...] Appends a new entry to a stream
- [x] XTRIM key MAXLEN [~] count Trims the stream to (approximately if '~' is passed) a certain size
- [x] XDEL key ID [ID ...] Removes the specified entries from the stream. Returns the number of items actually deleted, that may be different from the number of IDs passed in case certain IDs do not exist.
- [x] XRANGE key start end [COUNT count] Return a range of elements in a stream, with IDs matching the specified IDs interval
- [x] XREVRANGE key end start [COUNT count] Return a range of elements in a stream, with IDs matching the specified IDs interval, in reverse order (from greater to smaller IDs) compared to XRANGE
- [x] XLEN key Return the number of entires in a stream
- [x] XREAD [COUNT count] [BLOCK milliseconds] STREAMS key [key ...] id [id ...] Return never seen elements in multiple streams, with IDs greater than the ones reported by the caller for each stream. Can block.
- [x] XGROUP [CREATE key groupname id-or-$] [SETID key groupname id-or-$] [DESTROY key groupname] [DELCONSUMER key groupname consumername] Create, destroy, and manage consumer groups.
- [x] XREADGROUP GROUP group consumer [COUNT count] [BLOCK milliseconds] [NOACK] STREAMS key [key ...] ID [ID ...] Return new entries from a stream using a consumer group, or access the history of the pending entries for a given consumer. Can block.
- [x] XACK key group ID [ID ...] Marks a pending message as correctly processed, effectively removing it from the pending entries list of the consumer group. Return value of the command is the number of messages successfully acknowledged, that is, the IDs we were actually able to resolve in the PEL.
- [x] XCLAIM key group consumer min-idle-time ID [ID ...] [IDLE ms] [TIME ms-unix-time] [RETRYCOUNT count] [FORCE] [JUSTID] Changes (or acquires) ownership of a message in a consumer group, as if the message was delivered to the specified consumer.
- [x] XPENDING key group [start end count] [consumer] Return information and entries from a stream consumer group pending entries list, that are messages fetched but never acknowledged.

### Tips

anovakovic01:
This PR resolves part of the #157.

**Summary:**
- Add all of the commands except `XINFO`,
- Add new output types for commands like `XPENDING`,
- Add new input types,
- Fix `SET` command output type.

**Note:** Tests will be added in some other PR.

### DIFF
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -4,7 +4,6 @@ import java.time.Instant
 import java.util.concurrent.TimeUnit
 
 import scala.util.matching.Regex
-
 import zio.Chunk
 import zio.duration.Duration
 
@@ -273,10 +272,10 @@ object Input {
       Chunk(stringEncode("BLOCK"), stringEncode(data.value.toMillis.toString))
   }
 
-  case object StreamsInput extends Input[Map[String, String]] {
-    def encode(data: Map[String, String]): Chunk[RespValue.BulkString] = {
-      val keys = data.keys.foldLeft(Chunk.empty[RespValue.BulkString])((acc, key) => acc ++ stringEncode(key))
-      val ids  = data.values.foldLeft(Chunk.empty[RespValue.BulkString])((acc, id) => acc ++ stringEncode(id))
+  case object StreamsInput extends Input[((String, String), Chunk[(String, String)])] {
+    def encode(data: ((String, String), Chunk[(String, String)])): Chunk[RespValue.BulkString] = {
+      val (keys, ids) =
+        Chunk.fromIterable(data._1 +: data._2).map(pair => (stringEncode(pair._1), stringEncode(pair._2))).unzip
 
       Chunk.single(stringEncode("STREAMS")) ++ keys ++ ids
     }
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -4,6 +4,7 @@ import java.time.Instant
 import java.util.concurrent.TimeUnit
 
 import scala.util.matching.Regex
+
 import zio.Chunk
 import zio.duration.Duration
 
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -211,6 +211,99 @@ object Input {
       )
   }
 
+  case object IdleInput extends Input[Idle] {
+    def encode(data: Idle): Chunk[RespValue.BulkString] =
+      Chunk(stringEncode("IDLE"), stringEncode(data.millis.toString))
+  }
+
+  case object TimeInput extends Input[Time] {
+    def encode(data: Time): Chunk[RespValue.BulkString] =
+      Chunk(stringEncode("TIME"), stringEncode(data.millis.toString))
+  }
+
+  case object RetryCountInput extends Input[RetryCount] {
+    def encode(data: RetryCount): Chunk[RespValue.BulkString] =
+      Chunk(stringEncode("RETRYCOUNT"), stringEncode(data.count.toString))
+  }
+
+  case object XGroupCreateInput extends Input[XGroupCommand.Create] {
+    def encode(data: XGroupCommand.Create): Chunk[RespValue.BulkString] = {
+      val chunk =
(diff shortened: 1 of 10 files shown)


redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -211,6 +211,99 @@ object Input {
       )
   }
 
+  case object IdleInput extends Input[Idle] {
+    def encode(data: Idle): Chunk[RespValue.BulkString] =
+      Chunk(stringEncode("IDLE"), stringEncode(data.millis.toString))
+  }
+
+  case object TimeInput extends Input[Time] {
```
quelgar, line 219: Can this be done as an `Input[Duration]`?

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -211,6 +211,99 @@ object Input {
...
+    def encode(data: Time): Chunk[RespValue.BulkString] =
+      Chunk(stringEncode("TIME"), stringEncode(data.millis.toString))
+  }
+
+  case object RetryCountInput extends Input[RetryCount] {
```
quelgar, line 224: Similarly, `Input[Int]` seems fine?

redis/src/main/scala/zio/redis/Input.scala
anovakovic01: You're right. `IdleInput` too. There is a lot of stuff like this all over the project left from before usability PR was merged, so it will probably require separate PR to resolve it. I will check other input types related to streams too and remove what's unnecessary here.

redis/src/main/scala/zio/redis/Input.scala
anovakovic01: Yes, I'll remove it.

redis/src/main/scala/zio/redis/options/Streams.scala
```diff
@@ -0,0 +1,65 @@
...
+  case object WithJustId {
+    private[redis] def stringify = "JUSTID"
+  }
+
+  type WithJustId = WithJustId.type
+
+  sealed trait XGroupCommand
+
+  object XGroupCommand {
+    case class Create(key: String, group: String, id: String, mkStream: Boolean) extends XGroupCommand
```
mijicd, line 27: Make case classes final.

redis/src/main/scala/zio/redis/options/Streams.scala
```diff
@@ -0,0 +1,65 @@
...
+  )
+
+  case class PendingMessage(
+    id: String,
+    owner: String,
+    lastDelivered: Duration,
+    counter: Long
+  )
+
+  case class Block(value: Duration)
```
mijicd, line 54: These should probably be scraped.

mijicd (commented):
Let's address the "low hanging fruits" and merge. It can be "experimental" until we add some tests and play with it a bit.

redis/src/main/scala/zio/redis/options/Streams.scala
anovakovic01: This won't work for older Scala versions.

redis/src/main/scala/zio/redis/api/Streams.scala
```diff
@@ -0,0 +1,277 @@
...
+import zio.redis.Input._
+import zio.redis.Output._
+import zio.redis._
+import zio.{ Chunk, ZIO }
+
+trait Streams {
+  import Streams._
+  import XGroupCommand._
+
+  final def xAck(key: String, group: String, id: String, ids: String*): ZIO[RedisExecutor, RedisError, Long] =
```
mijicd, line 14: Add some scala doc either now or in the follow up with tests.

mijicd (approved):
Looks good for first phase, we can refine it later.

redis/src/main/scala/zio/redis/api/Streams.scala
anovakovic01: I will add it with tests, as there can be some API changes while writing tests.

mijicd (approved):

Issue #196, opened 2020-11-21
Add sets api to test executor


mschuwalow:
The [test executor](https://github.com/zio/zio-redis/blob/master/redis/src/main/scala/zio/redis/Interpreter.scala#L80) currently only supports a subset of the api.
The goal of this issue is to support the full sets api in the test executor. Tests for this group should be enabled for the test executor [here](https://github.com/zio/zio-redis/blob/master/redis/src/test/scala/zio/redis/ApiSpec.scala#L32)

mschuwalow:
I'm taking this one

mschuwalow:
Most of the API is done, feel free to do the remaining ones.

For [SRANDMEMBER](https://redis.io/commands/srandmember) we probably want something similar to [TRandom](https://github.com/zio/zio-keeper/blob/master/keeper/src/main/scala/zio/keeper/hyparview/TRandom.scala) in keeper

helanto:
@mschuwalow I can try to implement the remaining parts

PR #204, opened 2020-11-21
Partial support for sets api in test executor

mschuwalow:
part of #196

### DIFF
--- a/redis/src/main/scala/zio/redis/Interpreter.scala
+++ b/redis/src/main/scala/zio/redis/Interpreter.scala
@@ -74,8 +74,11 @@ trait Interpreter {
           .tapError(e => logger.error(s"Executor exiting: $e"))
     }
 
-    private final class InMemory() extends Service {
-      def execute(command: Chunk[RespValue.BulkString]): zio.IO[RedisError, RespValue] =
+    private final class Test(
+      sets: TMap[String, Set[String]],
+      strings: TMap[String, String]
+    ) extends Service {
+      override def execute(command: Chunk[RespValue.BulkString]): zio.IO[RedisError, RespValue] =
         for {
           name   <- ZIO.fromOption(command.headOption).orElseFail(ProtocolError("Malformed command."))
           result <- runCommand(name.asString, command.tail).commit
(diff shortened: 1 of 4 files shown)


redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -91,25 +99,194 @@ trait Interpreter {
...
-          case api.Connection.Ping.name =>
+          case api.Connection.Ping.name  =>
             STM.succeedNow {
               if (input.isEmpty)
                 RespValue.bulkString("PONG")
               else
                 input.head
             }
+          case api.Sets.SAdd.name        =>
+            val key = input.head.asString
```
mijicd, line 122: I'd use `headOption` and handle the empty case.

redis/src/test/scala/zio/redis/SetsSpec.scala
```diff
@@ -41,6 +41,7 @@ trait SetsSpec extends BaseSpec {
             value <- uuid
             _     <- set(key, value)
             added <- sAdd(key, "hello").either
+            //_      = println(added)
```
mijicd, line 44: :cough:

redis/src/test/scala/zio/redis/BaseSpec.scala
```diff
@@ -13,4 +14,7 @@ trait BaseSpec extends DefaultRunnableSpec {
   def instantOf(millis: Long): UIO[Instant] = UIO(Instant.now().plusMillis(millis))
 
   val uuid = UIO(UUID.randomUUID().toString)
+
+  val TestExecutorUnsupportedTag   = "test executor unsupported"
+  lazy val testExecutorUnsupported = tag(TestExecutorUnsupportedTag)
```
mijicd, line 19: Nice!

redis/src/test/scala/zio/redis/SetsSpec.scala
mschuwalow: fixed :D

redis/src/main/scala/zio/redis/Interpreter.scala
mschuwalow: I think it doesn't make sense here.
The api defines that there has to be at least one argument. If one isn't present it really is a programming / unrecoverable error. The only alternative that makes sense here IMO is to die the fiber, but this will achieve that too.

mijicd (approved):

Issue #177, opened 2020-11-21
Document sets API


mijicd:
The current sets API (see `zio.redis.api.Sets`) exposes functions with argument names that are generic and meaningless. I propose to rename those to match those used in the [official documentation](https://redis.io/commands). Additionally, it would be great to provide a small, single-sentence scaladoc for them.

justinhj:
I have this one in progress

PR #206, opened 2020-11-21
Update Sets API documentation with standard naming

justinhj:
This updates the naming of parameters where there is a first item and a list of further items. Based on recommendation by @anovakovic01 in PR #183

### DIFF
--- a/redis/src/main/scala/zio/redis/api/Sets.scala
+++ b/redis/src/main/scala/zio/redis/api/Sets.scala
@@ -13,12 +13,12 @@ trait Sets {
   /** Add one or more members to a set
    *
     * @param key Key of set to add to
-   * @param firstMember first member to add
-   * @param restMembers subsequent members to add
+   * @param member first member to add
+   * @param members subsequent members to add
    * @return Returns the number of elements that were added to the set, not including all the elements already present into the set
    */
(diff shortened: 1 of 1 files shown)


justinhj:
fixes #177

anovakovic01 (changes requested):
As @mijicd suggested, we decided to update scaladoc format so that it contains an empty line as a first line, so while you're at it, can you fix that as well?

justinhj:
`  /** Add one or more members to a set
   *
   * @param key Key of set to add to`

Not sure what you mean. Could you reformat this example in the desired format?

Also I notice that the scala format command is

mijicd:
@justinhj He meant:

```
/**
 * Add one or more ...
 */
```

anovakovic01:
> ```
>  /** Add one or more members to a set
>    *
>    * @param key Key of set to add to
> ```
> 
> Not sure what you mean. Could you reformat this example in the desired format?

Sure:
```
/** 
 *  Add one or more members to a set
 *
 *  @param key Key of set to add to
```

justinhj:
ok sure, I prefer that myself

mijicd (approved):

anovakovic01 (approved):

PR #217, opened 2020-11-22
Added Documentation for keys API 

### LINKED ISSUE #171: Document keys API
The current keys API (see `zio.redis.api.Keys`) exposes functions with argument names that are generic and meaningless. I propose to rename those to match those used in the [official documentation](https://redis.io/commands). Additionally, it would be great to provide a small, single-sentence scaladoc for them.

gagandeepkalra:
resolves #171

### DIFF
--- a/redis/src/main/scala/zio/redis/api/Keys.scala
+++ b/redis/src/main/scala/zio/redis/api/Keys.scala
@@ -13,72 +13,288 @@ import zio.{ Chunk, ZIO }
 trait Keys {
   import Keys.{ Keys => _, _ }
 
-  final def del(a: String, as: String*): ZIO[RedisExecutor, RedisError, Long] = Del.run((a, as.toList))
+  /**
+   * Removes the specified keys. A key is ignored if it does not exist.
 
-  final def dump(a: String): ZIO[RedisExecutor, RedisError, Chunk[Byte]] = Dump.run(a)
+   *
+   * @param key one required key
+   * @param keys maybe rest of the keys
+   * @return The number of keys that were removed.
+   *
+   * @see [[unlink]]
+   * @see <a href="https://redis.io/commands/del">https://redis.io/commands/del</a>
+   */
+  final def del(key: String, keys: String*): ZIO[RedisExecutor, RedisError, Long] = Del.run((key, keys.toList))
 
-  final def exists(a: String, as: String*): ZIO[RedisExecutor, RedisError, Boolean] = Exists.run((a, as.toList))
+  /**
+   * Serialize the value stored at key in a Redis-specific format and return it to the user.
+   *
+   * @param key key
+   * @return bytes for value stored at key
+   *
+   * @see <a href="https://redis.io/commands/dump">https://redis.io/commands/dump</a>
+   */
+  final def dump(key: String): ZIO[RedisExecutor, RedisError, Chunk[Byte]] = Dump.run(key)
 
-  final def expire(a: String, b: Duration): ZIO[RedisExecutor, RedisError, Boolean] = Expire.run((a, b))
+  /**
+   * The number of keys existing among the ones specified as arguments. Keys mentioned multiple times and existing are
+   * counted multiple times.
+   *
+   * @param key one required key
+   * @param keys maybe rest of the keys
+   * @return The number of keys existing.
+   *
+   * @see <a href="https://redis.io/commands/exists">https://redis.io/commands/exists</a>
+   */
+  final def exists(key: String, keys: String*): ZIO[RedisExecutor, RedisError, Boolean] = Exists.run((key, keys.toList))
 
-  final def expireAt(a: String, b: Instant): ZIO[RedisExecutor, RedisError, Boolean] = ExpireAt.run((a, b))
+  /**
+   * Set a timeout on key. After the timeout has expired, the key will automatically be deleted.
+   *
+   * @param key key
+   * @param timeout timeout
+   * @return true, if the timeout was set, false if the key didn't exist
+   *
+   * @see [[expireAt]]
+   * @see <a href="https://redis.io/commands/expire">https://redis.io/commands/expire</a>
+   */
+  final def expire(key: String, timeout: Duration): ZIO[RedisExecutor, RedisError, Boolean] = Expire.run((key, timeout))
 
-  final def keys(a: String): ZIO[RedisExecutor, RedisError, Chunk[String]] = Keys.Keys.run(a)
+  /**
+   * Deletes the key at the specific timestamp. A timestamp in the past will delete the key immediately.
+   *
+   * @param key key
+   * @param timestamp an absolute Unix timestamp (seconds since January 1, 1970)
+   * @return true, if the timeout was set, false if the key didn't exist
+   *
+   * @see [[expire]]
+   * @see <a href="https://redis.io/commands/expireat">https://redis.io/commands/expireat</a>
+   */
+  final def expireAt(key: String, timestamp: Instant): ZIO[RedisExecutor, RedisError, Boolean] = ExpireAt.run((key, timestamp))
 
+  /**
+   * Returns all keys matching pattern.
+   *
+   * @param pattern string pattern
+   * @return keys matching pattern
+   *
+   * @see <a href="https://redis.io/commands/keys">https://redis.io/commands/keys</a>
+   */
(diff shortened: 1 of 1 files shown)


redis/src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -13,72 +13,293 @@ import zio.{ Chunk, ZIO }
...
+   * This command blocks the current client until all the previous write commands are successfully transferred and acknowledged
+   * by at least the specified number of replicas.
+   *
+   * @param replicas minimum replicas to reach
+   * @param timeout 0 means to block forever.
+   * @return the number of replicas reached both in case of failure and success
+   *
+   * @see <a href="https://redis.io/commands/wait">https://redis.io/commands/wait</a>
+   */
+  final def wait_(replicas: Long, timeout: Long): ZIO[RedisExecutor, RedisError, Long] = Wait.run((replicas, timeout))
```
gagandeepkalra, line 302: possibly rename to \`wait\`

redis/src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -13,72 +13,293 @@ import zio.{ Chunk, ZIO }
...
+   * The number of keys existing among the ones specified as arguments. Keys mentioned multiple times and existing are
+   * counted multiple times.
+   *
+   * @param key one required key
+   * @param keys maybe rest of the keys
+   * @return The number of keys existing.
+   *
+   * @see <a href="https://redis.io/commands/exists">https://redis.io/commands/exists</a>
+   */
+  final def exists(key: String, keys: String*): ZIO[RedisExecutor, RedisError, Boolean] = Exists.run((key, keys.toList))
```
gagandeepkalra, line 49: Exists return type should be Int

redis/src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -13,72 +13,293 @@ import zio.{ Chunk, ZIO }
...
+
+  /**
+   * Returns the string representation of the type of the value stored at key.
+   *
+   * @param key key
+   * @return type of the value stored at key
+   *
+   * @see <a href="https://redis.io/commands/type">https://redis.io/commands/type</a>
+   */
+  final def typeOf(key: String): ZIO[RedisExecutor, RedisError, RedisType] = TypeOf.run(key)
```
gagandeepkalra, line 277: [doc](https://redis.io/commands/type) says there's a `stream` type, which is somehow  missing

redis/src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -13,72 +13,293 @@ import zio.{ Chunk, ZIO }
...
-
-  final def pExpireAt(a: String, b: Instant): ZIO[RedisExecutor, RedisError, Boolean] = PExpireAt.run((a, b))
-
-  final def pTtl(a: String): ZIO[RedisExecutor, RedisError, Duration] = PTtl.run(a)
-
+    host: String,
+    port: Long,
+    key: String,
+    destinationDb: Long,
+    timeout: Long,
```
gagandeepkalra, line 108: possibly change to duration

redis/src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -13,72 +13,293 @@ import zio.{ Chunk, ZIO }
...
+   * @see <a href="https://redis.io/commands/unlink">https://redis.io/commands/unlink</a>
+   */
+  final def unlink(key: String, keys: String*): ZIO[RedisExecutor, RedisError, Long] = Unlink.run((key, keys.toList))
+
+  /**
+   * This command blocks the current client until all the previous write commands are successfully transferred and acknowledged
+   * by at least the specified number of replicas.
+   *
+   * @param replicas minimum replicas to reach
+   * @param timeout 0 means to block forever.
```
gagandeepkalra: possibly change timout's type to duration

gagandeepkalra (commented):
While writing the documentation I came upon a few possible improvements, these were out of scope for the current PR; have commented instead.

Lastly [sort](https://redis.io/commands/sort) command is missing.

redis/src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -13,72 +13,293 @@ import zio.{ Chunk, ZIO }
...
+  /**
+   * Removes the specified keys. A key is ignored if it does not exist.
+
+   *
+   * @param key one required key
+   * @param keys maybe rest of the keys
+   * @return The number of keys that were removed.
+   *
+   * @see [[unlink]]
+   * @see <a href="https://redis.io/commands/del">https://redis.io/commands/del</a>
```
mijicd, line 25: I would remove this for the sake of consistency with other methods.

redis/src/main/scala/zio/redis/api/Keys.scala
mijicd: Well, it is according to the docs, but indicates whether or not the key exists, which IMHO should be encoded as boolean.

redis/src/main/scala/zio/redis/api/Keys.scala
mijicd: This might be a good idea. I'll double check and we can have it as a separate issue. Wdyt?

redis/src/main/scala/zio/redis/api/Keys.scala
mijicd: That's added since the initial version was implemented. We can file this as a separate issue. Wdyt?

redis/src/main/scala/zio/redis/api/Keys.scala
mijicd: It collides with existing wait.

mijicd (changes requested):
A few comments from my side. Generally speaking, remove the links to redis docs and it is good to go. Nice work!

redis/src/main/scala/zio/redis/api/Keys.scala
gagandeepkalra: So, when there are multiple keys, `exists` command can return an integer greater than 1, in that case our code would throw, failing to convert to Boolean. I'd think that we either change the type to Int or possibly drop varargs support.

      case object BoolOutput extends Output[Boolean] {

        override protected def tryDecode(respValue: RespValue): Boolean =
          respValue match {
            case RespValue.Integer(0) => false
            case RespValue.Integer(1) => true
            case other                => throw ProtocolError(s"$other isn't a boolean")
          }

      }

redis/src/main/scala/zio/redis/api/Keys.scala
gagandeepkalra: Sorry, I meant with the quotes on

redis/src/main/scala/zio/redis/api/Keys.scala
mijicd: Ahh, forgot about multikey. Please add this to the "tickets to open", and we'll fix it there. We can have a chunk of booleans instead.

mijicd: Ahh, forgot about multikey. Please add this to the "tickets to open", and we'll fix it there. We can have a chunk of booleans instead

redis/src/main/scala/zio/redis/api/Keys.scala
mijicd: I wouldn't do that, at least for now. Quoted functions are a bit awkward IMHO.

mijicd (commented):
Just don't forget to file the issues :)

redis/src/main/scala/zio/redis/api/Keys.scala
gagandeepkalra: sure 😄

mijicd (approved):

redis/src/main/scala/zio/redis/api/Keys.scala
gagandeepkalra: Picking up this thread again, if you could elaborate what you meant by chunk of booleans? I'd guess you mean a boolean for each key, but from the api we get a single integer for all.

mijicd: Ah, I see. Ignore the chunk part, just return integer.

gagandeepkalra: Sure, I've raised the review now.

PR #223, opened 2020-11-22
Sets API to test executor

### LINKED ISSUE #196: Add sets api to test executor

helanto:
closes #196

### DIFF
--- a/redis/src/main/scala/zio/redis/Interpreter.scala
+++ b/redis/src/main/scala/zio/redis/Interpreter.scala
@@ -225,6 +225,20 @@ trait Interpreter {
               sets.get(key).map(_.fold(Replies.EmptyArray)(Replies.array(_))),
               STM.succeedNow(Replies.WrongType)
             )
+          case api.Sets.SRem.name        =>
+            val key = input.head.asString
+            STM.ifM(isSet(key))(
+              {
+                val values = input.tail.map(_.asString)
+                for {
+                  oldSet <- sets.getOrElse(key, Set.empty)
+                  newSet  = oldSet -- values
+                  removed = oldSet.size - newSet.size
+                  _      <- sets.put(key, newSet)
+                } yield RespValue.Integer(removed.toLong)
+              },
+              STM.succeedNow(Replies.WrongType)
+            )
           case api.Strings.Set.name      =>
             // not a full implementation. Just enough to make set tests work
             val key   = input.head.asString
--- a/redis/src/main/scala/zio/redis/Interpreter.scala
+++ b/redis/src/main/scala/zio/redis/Interpreter.scala
@@ -251,6 +251,21 @@ trait Interpreter {
                 .map(unionSet => Replies.array(unionSet)),
               STM.succeedNow(Replies.WrongType)
             )
+          case api.Sets.SUnionStore.name =>
+            val destination = input.head.asString
+            val keys        = input.tail.map(_.asString)
+            STM.ifM(forAll(keys)(isSet))(
+              for {
+                union <- STM
+                           .foldLeft(keys)(Set.empty[String]) { (unionSoFar, nextKey) =>
+                             sets.getOrElse(nextKey, Set.empty[String]).map { currentSet =>
+                               unionSoFar ++ currentSet
+                             }
+                           }
+                _     <- sets.put(destination, union)
+              } yield RespValue.Integer(union.size.toLong),
+              STM.succeedNow(Replies.WrongType)
+            )
           case api.Sets.SScan.name       =>
             def maybeGetCount(key: RespValue.BulkString, value: RespValue.BulkString): Option[Int] =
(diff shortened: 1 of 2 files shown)


redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -29,15 +31,26 @@ trait Interpreter {
...
       (ZLayer.identity[Logging] ++ ByteStream.loopback(port)) >>> StreamedExecutor
 
-    val test: ZLayer[Any, Nothing, RedisExecutor] =
+    val test: ZLayer[zio.random.Random, Nothing, RedisExecutor] = {
+      val makePickRandom: URIO[zio.random.Random, Int => USTM[Int]] =
+        for {
+          seed   <- random.nextInt
+          sRandom = new scala.util.Random(seed)
+          ref    <- TRef.make(LazyList.continually((i: Int) => sRandom.nextInt(i))).commit
+        } yield (i: Int) => ref.modify(s => (s.head(i), s.tail))
```
mijicd, line 40: @mschuwalow will provide TRandom. However, until then, I propose to remove random generator completely, and use something not random at all, i.e. long circular buffer of elements.

@mschuwalow wdyt?

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -225,6 +239,99 @@ trait Interpreter {
...
+                           case Some(n) if n < 0 =>
+                             selectNWithReplacement(asVector, -1 * n, randomPick).map(Replies.array)
+                           case Some(0)          => STM.succeedNow(RespValue.NullValue)
+                         }
+                } yield res
+              },
+              STM.succeedNow(Replies.WrongType)
+            )
+          case api.Sets.SRem.name        =>
+            val key = input.head.asString
```
mijicd, line 265: I presume we're safe here as the client doesn't allow chunks to be empty, but please double check that.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -225,6 +239,99 @@ trait Interpreter {
...
+                _     <- sets.put(destination, union)
+              } yield RespValue.Integer(union.size.toLong),
+              STM.succeedNow(Replies.WrongType)
+            )
+          case api.Sets.SScan.name       =>
+            def maybeGetCount(key: RespValue.BulkString, value: RespValue.BulkString): Option[Int] =
+              key.asString match {
+                case "COUNT" => Some(value.asString.toInt)
+                case _       => None
+              }
```
mijicd, line 310: This will be more reliable than the actual count :).

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -225,6 +239,99 @@ trait Interpreter {
...
+                }
+                else None
+                val maybeCount =
+                  if (input.size > 4) maybeGetCount(input(4), input(5))
+                  else if (input.size > 2) maybeGetCount(input(2), input(3))
+                  else None
+                val end        = start + maybeCount.getOrElse(10)
+                for {
+                  set      <- sets.getOrElse(key, Set.empty)
+                  filtered  = maybeRegex.map(regex => set.filter(s => regex.pattern.matcher(s).matches)).getOrElse(set)
```
mijicd, line 327: I prefer fold to map and getOrElse.

redis/src/main/scala/zio/redis/Interpreter.scala
```diff
@@ -240,6 +347,33 @@ trait Interpreter {
           isString <- strings.contains(name)
         } yield !isString
 
+      private[this] def selectN[A](values: Vector[A], n: Long, pickRandom: Int => USTM[Int]): USTM[List[A]] = {
+        def go(remaining: Vector[A], toPick: Long, acc: List[A]): USTM[List[A]] =
+          (remaining, toPick) match {
+            case (Vector(), _) | (_, 0) => STM.succeed(acc)
```
mijicd, line 353: You can use succeedNow, it's internal, but handy.

mijicd (commented):
Generally looks good. @mschuwalow please take a look.

redis/src/main/scala/zio/redis/Interpreter.scala
mschuwalow: I think it's fine for now. It's already implemented and we can update it once we have a proper TRandom

mschuwalow (approved):

PR #227, opened 2020-11-24
WIP: Implement STRALGO API

rfmejia:


redis/src/main/scala/zio/redis/options/Strings.scala
```diff
@@ -64,4 +64,53 @@ trait Strings {
...
+    case class LongestCommonSubsequence(
+      lcsType: LcsType,
+      input1: String,
+      input2: String,
+      len: Option[QueryType] = None,
+      minMatchLen: Option[MinMatchLen] = None,
+      withMatchLen: Option[WithMatchLen] = None
+    ) extends StrAlgoCommand
+
+    sealed trait LcsType { self =>
```
anovakovic01, line 80: Maybe move these out of the `StrAlgoCommand` object and just leave `LongestCommonSubsequence` in it.

redis/src/main/scala/zio/redis/api/Strings.scala
```diff
@@ -254,6 +254,18 @@ trait Strings {
...
     SetRange.run((key, offset, value))
 
+  /**
+   * Implements complex algorithms that operate on strings. Right now the only algorithm implemented is
+   * the longest common subsequence.
+   *
+   * @param key Key of the string to get the length of
+   * @return Returns the length of the string
+   */
+  final def strAlgo(command: StrAlgoCommand): ZIO[RedisExecutor, RedisError, LongestCommonSubsequenceResult] =
```
anovakovic01, line 264: I recommend that you create `strAlgoLcs` function that will receive actual arguments. Also, when new options are added, the output will probably be of a different type, so it will be easier to create new functions.

rfmejia: Got it, wasn't sure what kind of API we would be best. So instead we should use the following:

`final def strAlgoLcs(lcsType: LcsType,
      input1: String,
      input2: String,
      len: Option[QueryType] = None,
      minMatchLen: Option[MinMatchLen] = None,
      withMatchLen: Option[WithMatchLen] = None
): ZIO[RedisExecutor, RedisError, LongestCommonSubsequenceResult]`

I can also remove `case class LongestCommonSubsequence()` and simplify its inputs accordingly. Does this seem reasonable?

rfmejia:
Note: This is still marked as WIP because of an issue with deserializing the output when using the "IDX" flag.

mijicd:
@rfmejia Do you mind if I close this draft as it seems inactive, and @barthorre started working on it in https://github.com/zio/zio-redis/pull/322?

rfmejia:
Sure, go ahead. Apologies but I couldn't make progress with the IDX issue. Hopefully the upstream changes to deserialization will make the new effort work.

PR #230, opened 2020-11-29
Change input type arguments from `Long` to `Duration` in key commands

justinhj:


docker-compose.yml
```diff
@@ -1,8 +1,14 @@
...
 services:
-  redis:
-    container_name: test_redis
+  redis1:
+    container_name: test_redis_1
     image: redis:6-alpine
     ports:
       - "6379:6379"
+
+  redis2:
```
justinhj, line 10: Note the general idea here is to test commands that require two Redis connections such as migrate, so the docker compose environment now brings up two servers with different exposed ports.

redis/src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -296,7 +296,7 @@ private[redis] object Keys {
         OptionalInput(AuthInput),
         OptionalInput(NonEmptyList(StringInput))
       ),
-      MultiStringOutput
+      StringOutput
```
justinhj, line 299: Fixes a bug

redis/src/test/scala/zio/redis/KeysSpec.scala
```diff
@@ -112,6 +119,48 @@ trait KeysSpec extends BaseSpec {
...
+            _         <- set(key, value)
+            response  <- migrate("redis1",
+                                  6379,
+                                  key,
+                                  0L,
+                                  JavaDuration.ofMillis(5000),
+                                  copy = Option(Copy),
+                                  replace = Option(Replace),
+                                  keys = None)
+            out       <- ZIO.access[SecondRedisExecutor](_.get.execute(Input.StringInput.encode("GET") ++ Input.StringInput.encode(key)))
```
justinhj, line 136: This almost works but it's a bit gross. It would be nice to use RedisCommand directly but they require a RedisExecutor not my new type. Maybe some implicit conversion could help?

redis/src/test/scala/zio/redis/SecondRedisExecutor.scala
```diff
@@ -0,0 +1,24 @@
+package zio.redis
+
+import zio.Has
+import zio.ZLayer
+import zio.Chunk
+import zio.logging.Logging
+
+object SecondRedisExecutorLayer {
```
justinhj, line 8: This is the main problem area. Trying to figure out how to make a new type of Redis Executor. Ideally, it would just reuse RedisExecutor but the class is quite complex and I'm having a hard time making it happen. @adamgfraser suggested using a case class, but not sure how. Perhaps as a simple wrapper? But then I'm not sure how to make that a ZLayer that can be threaded through the test.

redis/src/main/scala/zio/redis/RedisExecutor.scala
```diff
@@ -25,7 +26,7 @@ object RedisExecutor {
 
   private[this] final val RequestQueueSize = 16
 
-  private[this] final val StreamedExecutor =
+  final val StreamedExecutor =
```
justinhj, line 29: Made this public so I can use it in the second executor, maybe not a good idea

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -20,6 +24,15 @@ object ApiSpec
     with HashSpec
     with StreamsSpec {
 
+  case class RedisExecutorWrapper(re: RedisExecutor.Service)
```
justinhj, line 27: This is where I make the second redis executor by wrapping the service of the original

redis/src/test/scala/zio/redis/KeysSpec.scala
```diff
@@ -112,6 +125,50 @@ trait KeysSpec extends BaseSpec {
...
+            _         <- set(key, value)
+            response  <- migrate("redis2",
+                                  6379,
+                                  key,
+                                  0L,
+                                  JavaDuration.ofMillis(5000),
+                                  copy = Option(Copy),
+                                  replace = Option(Replace),
+                                  keys = None)
+            out       =  ZIO.accessM[RedisExecutorWrapper](_.re.execute(Input.StringInput.encode("GET") ++ Input.StringInput.encode(key)))
```
justinhj, line 142: One idea of what might work here is an implicit conversion from the wrapper to a RedisExecutor.Service

redis/src/test/scala/zio/redis/ApiSpec.scala
mijicd:
```suggestion
  final case class RedisExecutorWrapper(re: RedisExecutor.Service)
```

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -20,6 +24,15 @@ object ApiSpec
...
+
+  val secondConfigLayer = ZLayer.succeed(RedisConfig("localhost", 6380))
+
+  lazy val live: ZLayer[Logging with Has[RedisConfig], RedisError.IOError, Has[RedisExecutor.Service]] =
+    ZLayer.identity[Logging] ++ ByteStream.live >>> RedisExecutor.StreamedExecutor
```
mijicd, line 32: Can't you just pass a different configuration to the existing `RedisExecutor.live`?

redis/src/main/scala/zio/redis/RedisExecutor.scala
mijicd: I would keep it hidden.

redis/src/main/scala/zio/redis/RedisExecutor.scala
```diff
@@ -9,6 +9,7 @@ import zio.stream.Stream
 object RedisExecutor {
   trait Service {
     def execute(command: Chunk[RespValue.BulkString]): IO[RedisError, RespValue]
+    def test(n: Int) = n + 1
```
mijicd, line 12:
```suggestion
```

redis/src/test/scala/zio/redis/KeysSpec.scala
mijicd: Maybe it makes sense to use `get(key).provide(secondExecutor)`. Alternatively, you could group executors in the same fashion it's done on `TestClock`, so that the dependency of this test is `RedisExecutor with SecondRedisExecutor`.

redis/src/test/scala/zio/redis/ApiSpec.scala
justinhj: Maybe, will look into that

redis/src/test/scala/zio/redis/KeysSpec.scala
```diff
@@ -112,6 +134,27 @@ trait KeysSpec extends BaseSpec {
...
+            response  <- migrate("redis2",
+                                  6379,
+                                  key,
+                                  0L,
+                                  JavaDuration.ofMillis(5000),
+                                  copy = Option(Copy),
+                                  replace = Option(Replace),
+                                  keys = None)
+            value2    <- get(key).provide(secondRedisService)
+          } yield
```
justinhj, line 152: Trouble is this doesn't compile because secondRedisService is not of the right type for get
[error]  found   : zio.ZLayer[Any with Any,zio.redis.RedisError.IOError,zio.Has[RedisExecutorWrapper]]
[error]  required: zio.redis.RedisExecutor
[error]     (which expands to)  zio.Has[zio.redis.RedisExecutor.Service]
[error]             value2    <- get(key).provide(secondRedisService)
[error]                                           ^
[error] one error found

redis/src/test/scala/zio/redis/KeysSpec.scala
```diff
@@ -5,16 +5,38 @@ import zio.duration._
...
   ] with Has[TestRandom.Service] with Has[TestSystem.Service] with Has[RedisExecutor.Service] with Has[
     Annotations.Service
-  ], TestFailure[RedisError], TestSuccess] =
+  ], TestFailure[RedisError], TestSuccess] = {
+    final case class RedisExecutorWrapper(re: RedisExecutor.Service)
+
+    val secondConfigLayer = ZLayer.succeed(RedisConfig("localhost", 6380))
+    val secondRedisService = (Logging.ignore ++ secondConfigLayer >>> RedisExecutor.live).fresh.project(service => RedisExecutorWrapper(service))
+
+    // Implicit conversion from RedisExecutorWrapper to
```
justinhj, line 38: Maybe need some kind of implicit conversion from RedisExecutorWrapper to RedisExecutor, but not sure quite what it would look like.

redis/src/test/scala/zio/redis/KeysSpec.scala
```diff
@@ -5,16 +5,38 @@ import zio.duration._
...
   val keysSuite: Spec[Has[Clock.Service] with Has[RedisExecutor.Service] with Has[TestClock.Service] with Has[
     TestConsole.Service
   ] with Has[TestRandom.Service] with Has[TestSystem.Service] with Has[RedisExecutor.Service] with Has[
     Annotations.Service
-  ], TestFailure[RedisError], TestSuccess] =
+  ], TestFailure[RedisError], TestSuccess] = {
+    final case class RedisExecutorWrapper(re: RedisExecutor.Service)
+
+    val secondConfigLayer = ZLayer.succeed(RedisConfig("localhost", 6380))
+    val secondRedisService = (Logging.ignore ++ secondConfigLayer >>> RedisExecutor.live).fresh.project(service => RedisExecutorWrapper(service))
```
anovakovic01, line 36: You just need this:
```suggestion
    val secondRedisService = (Logging.ignore ++ secondConfigLayer) >>> RedisExecutor.live
```
Then, instead of `provide` use `provideLayer` method in your test and give it `secondRedisService`.

anovakovic01 (changes requested):

redis/src/test/scala/zio/redis/KeysSpec.scala
justinhj: Ah nice, thanks, still haven't figured out the differences between all the different provides :)

redis/src/test/scala/zio/redis/KeysSpec.scala
justinhj: Yeah provideLayer works

redis/src/test/scala/zio/redis/KeysSpec.scala
```diff
@@ -1,20 +1,31 @@
...
   ] with Has[TestRandom.Service] with Has[TestSystem.Service] with Has[RedisExecutor.Service] with Has[
     Annotations.Service
-  ], TestFailure[RedisError], TestSuccess] =
+  ], TestFailure[RedisError], TestSuccess] = {
+
+    // Note the migrate command tests require two Redis instances, so here we create a second
+    // Redis RedisExecutor
+    val secondConfigLayer  = ZLayer.succeed(RedisConfig("localhost", 6380))
+    val secondRedisService = (Logging.ignore ++ secondConfigLayer >>> RedisExecutor.live).fresh
+    val migrateTimeout     = JavaDuration.ofMillis(5000)
```
mijicd, line 27: I think you could use zio's duration here. Also, the constants can go to companion.

mijicd (approved):
Just a super small remark, but it can be fixed in a follow up. Other than that, looks good to me!

anovakovic01 (changes requested):

redis/src/test/scala/zio/redis/KeysSpec.scala
```diff
@@ -1,20 +1,28 @@
...
 import zio.duration._
+import zio.redis.RedisError.ProtocolError
 import zio.test.Assertion._
 import zio.test.TestAspect._
 import zio.test._
 import zio.test.environment.{ TestClock, TestConsole, TestRandom, TestSystem }
 import zio.{ Chunk, Has, ZIO }
 
+object KeysSpec {
+  val migrateTimeout: Duration = Duration.fromMillis(5000)
```
mijicd, line 15: Let's use zio's duration please. It delegates to java duration anyhow, but the syntax is a bit nicer.

anovakovic01 (approved):

PR #244, opened 2020-12-04
Fix Keys#exists to return Long instead of Boolean

### LINKED ISSUE #219: Fix `exists` command return type from the Keys api
When there are multiple keys, `exists` command can return an integer greater than 1, in that case our code would throw trying to interpret the result to a Boolean.

please refer this [thread](https://github.com/zio/zio-redis/pull/217#discussion_r528308091).

gagandeepkalra:
resolves #219


mschuwalow (approved):
Looks good :+1:

Issue #158, opened 2020-11-20
Implement scripting API


mijicd:
Extracted from [this comment](https://github.com/zio/zio-redis/issues/68#issuecomment-691777714) authored by @regis-leray.

### Commands

- [x] EVAL script numkeys key [key ...] arg [arg ...] Execute a Lua script server side
- [x] EVALSHA sha1 numkeys key [key ...] arg [arg ...] Execute a Lua script server side
- [ ] SCRIPT DEBUG YES|SYNC|NO Set the debug mode for executed scripts.
- [x] SCRIPT EXISTS sha1 [sha1 ...] Check existence of scripts in the script cache.
- [ ] SCRIPT FLUSH Remove all the scripts from the script cache.
- [ ] SCRIPT KILL Kill the script currently in execution.
- [x] SCRIPT LOAD script Load the specified Lua script into the script cache.

### Tips

- Define the API in `zio.redis.api`.
- Command-specific options should be defined in `zio.redis.options`.
- Additional inputs and outputs can be defined in `zio.redis.Input` and `zio.redis.Output`, respectively.
- Redis API evolves, therefore the commands listed above might be slightly changed. Double check [the official docs](https://redis.io/commands) before starting with implementation. 
- Use telnet to understand command protocol details.
- Write inputs and outputs tests in `zio.redis.InputSpec` and `zio.redis.OutputSpec`, respectively.
- Write integration tests by expanding the `zio.redis.ApiSpec`.
- If you feel that the pull request size is growing out of control, feel free to split it but make sure to link this issue in each of the related PRs.

mijicd:
Implemented by @anatolysergeev in #269

drmarjanovic:
@mijicd, can I work on this one?

mijicd:
All yours!

mijicd:
All yours!

Issue #214, opened 2020-11-22
API for binary data


quelgar:
There are many cases where the public API uses `String` where Redis supports arbitrary binary data. To use the Redis parlance, all bulk strings are "binary safe".

Taking the `SET` command for example, Redis supports arbitrary bytes for both the value and the key, meaning we should offer an API along the lines of:

```scala
final def set(
    key: Chunk[Byte],
    value: Chunk[Byte],
    expire: Option[Duration] = None,
    update: Option[Update] = None,
    keepTtl: Option[KeepTtl] = None
  ): ZIO[RedisExecutor, RedisError, Option[Unit]]
```

However, it will be extremely common in practice for users to want to use `String`, so we should continue to make this easy to do. How to best do this?

A simple brute-force method would be to offer two versions of all APIs where this matters, one using `Chunk[Byte]` and another using `String` (which simply passes UTF-8 encoded bytes to the `Chunk[Byte]` variant). I think that means a lot of boilerplate though, and it would not allow mixing and matching: it would be nice if we could do `set("myKey", myJpegBytes)`.

The best idea I've had so far is using typeclasses for encoding decoding. Encoding sketch:

```scala

trait RedisEncoder[A] {
  def encode(a: A): Chunk[Byte]
}

final def set[A: RedisEncoder, B: RedisEncoder](
    key: A,
    value: B,
    expire: Option[Duration] = None,
    update: Option[Update] = None,
    keepTtl: Option[KeepTtl] = None
  ): ZIO[RedisExecutor, RedisError, Option[Unit]] = {
  val keyBytes = implicitly[RedisEncoder[A]].encode(key)
  val valueBytes = implicitly[RedisEncoder[B]].encode(value)
  Set.run((keyBytes, valueBytes, expire, update, keepTtl)) 
}

implicit val identityEncoder: RedisEncoder[Chunk[Byte]] = new RedisEncoder[Chunk[Byte]] {
  def encode(a: Chunk[Byte]) = a
}

implicit val stringEncoder: RedisEncoder[String] = new RedisEncoder[String] {
  def encode(a: String) = Chunk.fromArray(s.getBytes(StandardCharsets.UTF_8))
}
```

mijicd:
I'm absolutely for it. The additional benefit - our API becomes less "stringly" typed. Typeclass approach is fine, we can provide the standard instances in companions and we'd still be adhering to that "have a single import" rule.

PR #269, opened 2021-01-08
Scripting API (eval, evalSha, scriptExists, scriptLoad)

anatolysergeev:
Issue #158

That is a try to implement scripting API for redis.
This PR have debatable changes: removing sealed state for traits Input and Output. I couldn't find an answer for questions "why they are should be sealed?", "why we don't want to give the opportunity for customers of this library to implement their own encoders and decoders".
In test I've written an example how that could be easy to implement own decoder for customer if traits wouldn't be sealed.
Thanks, I'm waiting for approval to continue coding in this approach or for comments how better it should be implemented.


redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -8,8 +8,8 @@ import scala.util.matching.Regex
 import zio.Chunk
 import zio.duration.Duration
 
-sealed trait Input[-A] {
-  private[redis] def encode(data: A): Chunk[RespValue.BulkString]
+trait Input[-A] {
```
mijicd, line 11: Please leave it sealed.

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -3,8 +3,7 @@ package zio.redis
 import zio.Chunk
 import zio.duration._
 
-sealed trait Output[+A] {
-
+trait Output[+A] {
```
mijicd, line 6: Please leave it sealed.

redis/src/main/scala/zio/redis/api/Scripting.scala
```diff
@@ -0,0 +1,18 @@
+package zio.redis.api
+
+import zio.ZIO
+import zio.redis.Input.EvalInput
+import zio.redis._
+import Scripting._
```
mijicd, line 6: Please move this to trait.

redis/src/main/scala/zio/redis/api/Scripting.scala
```diff
@@ -0,0 +1,18 @@
...
+
+trait Scripting {
+
+  def eval[K: Input, A: Input, R: Output](e: Script[K, A]): ZIO[RedisExecutor, RedisError, R] =
+    evalCommand[K, A, R].run(e)
+}
+
+private[redis] object Scripting {
+
+  final def evalCommand[K: Input, A: Input, R: Output]: RedisCommand[Script[K, A], R] =
```
mijicd, line 16: Please rename to `eval`. This will change as soon as you adjust input.

redis/src/main/scala/zio/redis/options/Scripting.scala
```diff
@@ -0,0 +1,6 @@
+package zio.redis.options
+
+trait Scripting {
+
+  case class Script[K, A](lua: String, keys: Seq[K], args: Seq[A])
```
mijicd, line 5: Let's not use `Chunk` instead of `Seq`. Also, make it sealed. More important, this isn't really an option.

redis/src/test/scala/zio/redis/ScriptingSpec.scala
```diff
@@ -0,0 +1,66 @@
...
+            arg1  <- uuid
+            arg2  <- uuid
+            lua    = """return {KEYS[1],KEYS[2],ARGV[1],ARGV[2]}"""
+            script = Script(lua, Seq(key1, key2), Seq(arg1, arg2))
+            res   <- eval(script)(StringInput, StringInput, MultiStringChunkOutput)
+          } yield assert(res)(equalTo(Chunk(key1, key2, arg1, arg2)))
+        },
+        testM("return custom data type") {
+
+          case class CustomData(count: Long, avg: Long, pair: (Int, String))
```
mijicd, line 29: Please make it final.

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -409,11 +409,19 @@ object Input {
     def encode(data: Update): Chunk[RespValue.BulkString] = Chunk.single(encodeString(data.stringify))
   }
 
-  final case class Varargs[-A](input: Input[A]) extends Input[Iterable[A]] {
+  final case class Varargs[-A](input: Input[A]) extends Input[Iterable[A]] { // todo: never used ??
     def encode(data: Iterable[A]): Chunk[RespValue.BulkString] =
       data.foldLeft(Chunk.empty: Chunk[RespValue.BulkString])((acc, a) => acc ++ input.encode(a))
   }
 
+  case class EvalInput[K, A](keysInput: Input[K], argsInput: Input[A]) extends Input[Script[K, A]] {
```
mijicd, line 417: I don't think this is correct. What you have in `Script` should be specified as an input, i.e.:
- string script
- varargs keys
- varargs args

Also, be consistent and make it final.

mijicd (changes requested):

redis/src/main/scala/zio/redis/Input.scala
anatolysergeev: Ok, add redisencoder, redisdecoder according to #214

redis/src/main/scala/zio/redis/options/Scripting.scala
anatolysergeev: Sorry, i don't understand that comment((

redis/src/main/scala/zio/redis/Input.scala
anatolysergeev: Hopefully, I understood you right and did how you wanted)

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -409,11 +411,21 @@ object Input {
     def encode(data: Update): Chunk[RespValue.BulkString] = Chunk.single(encodeString(data.stringify))
   }
 
-  final case class Varargs[-A](input: Input[A]) extends Input[Iterable[A]] {
+  final case class Varargs[-A](input: Input[A]) extends Input[Iterable[A]] { // todo: never used ??
     def encode(data: Iterable[A]): Chunk[RespValue.BulkString] =
       data.foldLeft(Chunk.empty: Chunk[RespValue.BulkString])((acc, a) => acc ++ input.encode(a))
   }
 
+  case object EvalInput extends Input[(String, Seq[Chunk[Byte]], Seq[Chunk[Byte]])] {
```
mijicd, line 419: Please no `Seq`.

redis/src/main/scala/zio/redis/options/Scripting.scala
```diff
@@ -0,0 +1,12 @@
+package zio.redis.options
+
+import zio.Chunk
+import zio.redis.RedisEncoder
+
+trait Scripting {
+
+  sealed case class Script[K, A](lua: String, keys: Seq[K], args: Seq[A]) {
```
mijicd, line 8: As mentioned in my previous comments, this isn't an option, and it isn't even necessary anymore.

redis/src/main/scala/zio/redis/options/Scripting.scala
mijicd: Sorry, my english broke, I wanted to say let's use `Chunk` instead of `Seq` and let's avoid using `Seq`.

redis/src/main/scala/zio/redis/RedisEncoder.scala
```diff
@@ -0,0 +1,29 @@
+package zio.redis
+
+import zio.Chunk
+
+import java.nio.charset.StandardCharsets
+
+trait RedisEncoder[A] {
```
mijicd, line 7: I'd rename to `Encoder`.

redis/src/main/scala/zio/redis/RedisDecoder.scala
```diff
@@ -0,0 +1,7 @@
+package zio.redis
+
+import zio.IO
+
+trait RedisDecoder[A] {
```
mijicd, line 5: I'd rename to `Decoder`.

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -36,6 +35,10 @@ object Output {
 
   import RedisError._
 
+  case object RespValueOutput extends Output[RespValue] {
```
mijicd, line 38: Let me think about this one a bit.

mijicd:
@quelgar this PR introduces the notion of codecs you mentioned in #214. Now, it doesn't propagates them throughout the API, but we might as well settle upon this API now and propagate in a separate ticket.

One thing I'm not sure is `Decoder` being aware of `RespValue`, but it might be the only way to encode it (too late for me to think).

redis/src/main/scala/zio/redis/options/Scripting.scala
anatolysergeev: Script has been deleted. IMHO this class showed relation between keys, args and lua script, that's why i wrote it)

redis/src/main/scala/zio/redis/Decoder.scala
```diff
@@ -0,0 +1,7 @@
+package zio.redis
+
+import zio.IO
+
+trait Decoder[A] {
```
mijicd, line 5: I am unsure about this. `RespValue` should not leak in the client code. If we're allowing custom decoders (and we should), I think they should work on byte chunks.

redis/src/main/scala/zio/redis/Encoder.scala
```diff
@@ -0,0 +1,29 @@
+package zio.redis
+
+import java.nio.charset.StandardCharsets
+
+import zio.Chunk
+
+trait Encoder[A] {
```
mijicd, line 7: Let's add summoner here.

redis/src/main/scala/zio/redis/Output.scala
mijicd: Updated the comment above.

redis/src/main/scala/zio/redis/api/Scripting.scala
```diff
@@ -0,0 +1,136 @@
...
+   * Evaluates a Lua script
+   *
+   * @param script Lua script
+   * @param keys keys available through KEYS param in the script
+   * @param args values available through ARGV param in the script
+   * @return redis protocol value that is converted from the Lua type.
+   *         You have to write decoder that would convert
+   *         redis protocol value to a suitable type for your app
+   *
+   * @since 2.6.0
```
mijicd, line 21: I'd remove `@since` as it refers to Redis version, which doesn't really matches the lib version.

redis/src/main/scala/zio/redis/api/Scripting.scala
```diff
@@ -0,0 +1,136 @@
...
+  def scriptLoad(script: String): ZIO[RedisExecutor, RedisError, String] =
+    ScriptLoad.run(script)
+}
+
+private[redis] object Scripting {
+
+  final val Eval: RedisCommand[(String, Chunk[Chunk[Byte]], Chunk[Chunk[Byte]]), RespValue] =
+    RedisCommand("EVAL", EvalInput, RespValueOutput)
+
+  final val EvalSHA: RedisCommand[(String, Chunk[Chunk[Byte]], Chunk[Chunk[Byte]]), RespValue] =
```
mijicd, line 118: Please rename to `EvalSha`.

redis/src/test/scala/zio/redis/ScriptingSpec.scala
```diff
@@ -0,0 +1,279 @@
+package zio.redis
+
+import zio._
+import zio.redis.RedisError._
+import zio.test.Assertion._
+import zio.test.TestAspect.ignore
+import zio.test._
+
+trait ScriptingSpec extends BaseSpec {
+  val scriptingSpec: Spec[Annotations with RedisExecutor, TestFailure[Any], TestSuccess] =
```
mijicd, line 10: Lots of tests seems to be ignored, what's the problem?

mijicd (commented):
@anatolysergeev Please run `sbt prepare` as well.

@quelgar Just a reminder to take a look at codec part :).

quelgar:
> @quelgar this PR introduces the notion of codecs you mentioned in #214. Now, it doesn't propagates them throughout the API, but we might as well settle upon this API now and propagate in a separate ticket.
> 
> One thing I'm not sure is `Decoder` being aware of `RespValue`, but it might be the only way to encode it (too late for me to think).

Yeah, I can't help thinking we're missing an opportunity to simplify here. The fact there's commands here that use the `RespValueOutput` no-op output, then does the decoding via a different type, seems like code smell. We're bypassing the `Output` mechanism altogether, and we'll potentially end up re-implementing decoding to common types like `Instant` and `Duration`.

Can we pull `Output` up to the API level as a type class? The unsafe methods are always private. I think user-provided implementations could all be built from the basic `Output` implementations, via `map`. We might want a variant of `map` to support failure.

anatolysergeev:
> Can we pull `Output` up to the API level as a type class? The unsafe methods are always private. I think user-provided implementations could all be built from the basic `Output` implementations, via `map`. We might want a variant of `map` to support failure.

That was my first implementation. But @mijicd proposed to do Encoder. And i rewrote Decoder too. Main points were consistency with Input  and save ADT for Output. I agree that for contributors to this library it would harder to understand why there are 2 "similar" services, but for users i don't think that would be a problem.

I have more worries about that or mb i don't understand something
> I am unsure about this. RespValue should not leak in the client code. If we're allowing custom decoders (and we should), I think they should work on byte chunks.
User should implement their own Resp deserialization or how that would work?
The problem is that eval could return any type of RespValue, and only user know what lua script would return

mijicd:
@anatolysergeev Well, the original implementation unsealed the output and exposed it in user-land. I don't think we should do it, and I recalled an issue that proposed encoder / decoder mechanism. If we're aiming to simplify, then let's propose combinators to the existing input / output as @quelgar wrote.

redis/src/test/scala/zio/redis/ScriptingSpec.scala
anatolysergeev: test for flash command need specific test environment to run, due to constancy i wrote ignore.
with other tests i was straggling to wrote them that they would work like i want. For example test for command kill i don;t understand why but don't want to kill the script((, test for debug when you start command yes or sync block evrething in redis and don't want to turn off debug mode. 
I think that debug, kill should be run to in specific environment.

anatolysergeev:
Ok. I want to do some conclusion here, what i am going to do
1. delete Encoder/Decoder
2. Add map combinator to Input (Output already has that one). i want remember u, that It would look like this:
image
3. sealed is stayed for Input and Output
4. add some catch method to Input and Output
5. RespValueOutput is stayed too because we need initial output of respvalue 

am i right?) It would be nice to have approve from both of you. Thanks.

P.S. hard decision actually i don't even know, what of this i like more)) Can i ask why for input and output sealed status so important?

mijicd:
@anatolysergeev i believe i answered that one. Originally, input and output where envisioned as command shapes. Since there's a finite amount of those, it makes sense to seal them. However, now we're effectively giving them more power. I'd say try to keep them sealed, if we need to unseal, we can do that.

redis/src/test/scala/zio/redis/ScriptingSpec.scala
mijicd: Hm, in that case i propose shipping the testable commands in this PR and the other through follow up pull requests, i.e. once the env is figured out.

anatolysergeev: Ok

quelgar:
@anatolysergeev sounds reasonable to me

I think for `Input` we want `contramap` instead of `map`:

```scala
sealed trait Input[A] {

  self =>

  final def contramap[B](f: B => A): Input[B] = new Input {
    def encode(data: B) = self.encode(f(b))
  }
}
```

On the point "`eval` can return any `RespValue`", I was wondering if for a particular script you'd know what type of value to expect back. But I guess it's all dynamically typed, so your script could return a bulk string on one code path and an integer on another. What kind of monster Scala programmer would write such a script? But we should support it regardless 🤣

anatolysergeev:
I haven't done methods that would be catch failures( 
I have been thinking how to write the methods and best what i can figured out is this:
``` 
final def catchSome[E <: RedisError, B >: A](rePF: PartialFunction[E, B]): Output[B] = {
 val thPF: PartialFunction[Throwable, B] = { th => throw th }
  new Output[B] {
    protected def tryDecode(respValue: RespValue): B = Try(self.tryDecode(respValue)).recover(rePF.orElse[Throwable, B](thPF)).get
  }
}
```

mijicd:
@anatolysergeev I am sorry for my way overdue response. My workload took its toll. After a discussion with John, I realized that `zio-schema` is close to release. That said, I don't think we should invent the new codec "machinery" but used the one provided by `zio-schema` instead. With that in mind, I propose the following course of actions:

- extract all encoding and decoding from this PR (stay in string domain) in a way that it provides a part of scripting commands implementable with the current design
- introduce the codecs from `zio-schema` in the follow-up
- propagate encoding changes throughout the library (this would involve changing the type constraints on our API level and postponing codec operations in the interpreter)

anatolysergeev:
Hello @mijicd . I am sorry for my long response too and my inactivity. I've read your comment and don't see solution with string domain for eval appropriate and didn't want to do it just to close the issue.
Has zio-schema been released? I saw some release tags there.
Could I have a try to write a solution with schema instead current encoders/decoders (input/output) approach? Or we want to remove Input/Output totally from this project in favor of zio.Schema and there are no way to do something for eval now?

mijicd:
@anatolysergeev Please don't remove input / output. As originally discussed, they're intended to specify shapes of commands. However, we need arbitrary input / output, that could be polymorphic. In the user land (API functions), this parameter would be type parameterized to require a schema. So yes, I propose digging into zio-schema API (it's been released), and drafting the solution. I do have a sketch of it, so if you're willing to wait until the weekend, we could have a pairing session. WDYT?

anatolysergeev:
I think that a cool idea and i very excited about that. But if you mean in "pair session" an online conversation i should warn you that i'm not confident in my English speaking skills by i would try my best))

anovakovic01 (approved):

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -557,6 +564,21 @@ object Input {
       data.foldLeft(Chunk.empty: Chunk[RespValue.BulkString])((acc, a) => acc ++ input.encode(a))
   }
 
+  final case class EvalInput[-K, -V](inputK: Input[K], inputV: Input[V]) extends Input[(String, Chunk[K], Chunk[V])] {
+    def encode(data: (String, Chunk[K], Chunk[V]))(implicit codec: Codec): Chunk[RespValue.BulkString] = {
+      val (lua, keys, args) = data
+      val encodedScript     = Chunk(encodeString(lua), encodeString(keys.size.toString))
+      val encodedKeys       = keys.foldLeft[Chunk[BulkString]](Chunk.empty)((cur, next) => cur ++ inputK.encode(next))
+      val encodedArgs       = args.foldLeft[Chunk[BulkString]](Chunk.empty)((cur, next) => cur ++ inputV.encode(next))
```
mijicd, line 572: Shouldn't
```suggestion
      val encodedKeys       = keys.flatMap(inputK.encode)
      val encodedArgs       = args.flatMap(inputV.encode)
```

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -591,4 +613,13 @@ object Input {
...
   }
+
+  trait Implicits {
+    implicit val bytesEncoder: Input[Chunk[Byte]] = ByteInput
+    implicit val booleanInput: Input[Boolean]     = BoolInput
+    implicit val stringInput: Input[String]       = StringInput
+    implicit val longInput: Input[Long]           = LongInput
+  }
+
+  object Implicits extends Implicits
```
mijicd, line 624: This seems to be used only in tests. If that's the case, please define them there instead.

redis/src/main/scala/zio/redis/api/Scripting.scala
```diff
@@ -0,0 +1,119 @@
...
+   *   redis protocol value that is converted from the Lua type. You have to write decoder that would convert redis
+   *   protocol value to a suitable type for your app
+   */
+  def eval[K: Input, A: Input](
+    script: String,
+    keys: Chunk[K],
+    args: Chunk[A]
+  ): ResultOutputBuilder = new ResultOutputBuilder {
+    def returning[R: Output]: ZIO[RedisExecutor, RedisError, R] = {
+      val command = RedisCommand(Eval, EvalInput(implicitly[Input[K]], implicitly[Input[A]]), implicitly[Output[R]])
```
mijicd, line 53: If possible, add summoners for Input and Output. It'll make the call site cleaner.

redis/src/test/scala/zio/redis/ScriptingSpec.scala
```diff
@@ -0,0 +1,260 @@
...
+
+  object CustomInputValue {
+    implicit val encoder: Input[CustomInputValue] = Input.Tuple2(StringInput, LongInput).contramap { civ =>
+      (civ.name, civ.age)
+    }
+  }
+
+  final case class CustomData(count: Long, avg: Long, pair: (Int, String))
+
+  object CustomData {
```
mijicd, line 212: I would not nest items not directly related to CustomData in its companion object.

redis/src/test/scala/zio/redis/ScriptingSpec.scala
```diff
@@ -0,0 +1,260 @@
...
+    private val tryDecodeLong: RespValue => Long = {
+      case RespValue.Integer(value) => value
+      case other                    => throw ProtocolError(s"$other isn't a integer type")
+    }
+    private val tryDecodeString: RespValue => String = {
+      case s @ RespValue.BulkString(_) => s.asString
+      case other                       => throw ProtocolError(s"$other isn't a string type")
+    }
```
mijicd, line 220: You already have these in LongOutput and StringOutput. Please check the rest implicits you defined here.

mijicd (changes requested):

redis/src/test/scala/zio/redis/ScriptingSpec.scala
anatolysergeev: Yes, that's true. But this outputs required codecs and we don't have schema codecs implicit in the `Output.map`. And i had a long thought about that, i can not figure out how better to handle this. 
1. Codecs are serialize, deserialize that we put into cache ignoring redis protocol, but here we work with redis protocol
2. It seems not very good to have initialization of codec in few place (now RedisExacutor has dependency of it)
3. It looks like we should do something with `.map` method because inside it we actually has required implicit of codec, but how to get it from there 🤔

anatolysergeev: One of idea was to add new `.mapWithCodec` - not sure about a name))
```scala
  final def mapWithCodec[B](f: Codec => A => B): Output[B] =
    new Output[B] {
      protected def tryDecode(respValue: RespValue)(implicit codec: Codec): B = f(codec)(self.tryDecode(respValue))
    }
```

mijicd (approved):

PR #294, opened 2021-02-26
add test for connection

jxnu-liguobin:


redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -27,6 +30,24 @@ private[redis] final class TestExecutor private (
...
+        STM.succeedNow {
+          if (input.isEmpty) errResponse(api.Connection.Echo.name)
+          else
+            input.head
+        }
+      case api.Connection.Select.name =>
+        STM.succeedNow {
+          if (input.isEmpty) errResponse(api.Connection.Select.name)
+          else RespValue.bulkString("OK")
+        }
```
mijicd, line 49:
```suggestion
      case api.Connection.Auth.name =>
        STM.succeedNow(if (input.isEmpty) errResponse(name) else RespValue.bulkString("OK"))
        
      case api.Connection.Echo.name =>
        STM.succeedNow(if (input.isEmpty) errResponse(name) else input.head)
        
      case api.Connection.Select.name =>
        STM.succeedNow(if (input.isEmpty) errResponse(name) else RespValue.bulkString("OK"))
```

Optionally, you can extract a helper.

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -4,29 +4,48 @@ import scala.collection.compat.immutable.LazyList
...
     name match {
       case api.Connection.Ping.name =>
-        STM.succeedNow {
-          if (input.isEmpty)
-            RespValue.bulkString("PONG")
-          else
-            input.head
-        }
-
+        succeedNowMayEmpty(emptyReturn = RespValue.bulkString("PONG"), nonEmptyReturn = input.head)
```
mijicd, line 36: Please don't name arguments.

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -4,29 +4,48 @@ import scala.collection.compat.immutable.LazyList
...
     for {
       name   <- ZIO.fromOption(command.headOption).orElseFail(ProtocolError("Malformed command."))
       result <- runCommand(name.asString, command.tail).commit
     } yield result
 
-  private[this] def runCommand(name: String, input: Chunk[RespValue.BulkString]): STM[RedisError, RespValue] =
+  private def errResponse(cmd: String): RespValue.BulkString =
+    RespValue.bulkString(s"(error) ERR wrong number of arguments for '$cmd' command")
+
+  private def succeedNowMayEmpty(emptyReturn: => BulkString, nonEmptyReturn: => BulkString)(implicit
```
mijicd, line 25: Let's make it a bit more API specific, and drop the implicit, so:

```suggestion
private def onConnection(command: String, input: Chunk[RespValue.BulkString])(res: => RespValue.BulkString) = STM.succeedNow(if (input.isEmpty) errResponse(command) else res)
```
 
Then you can propagate it to the call site. Alrernatively, you could make it curried in a way that it returna a function from input to result.

mijicd (commented):
One last change and it's good to go.

mijicd (approved):

PR #302, opened 2021-03-04
Support SORT command in Keys API

### LINKED ISSUE #218: Add `sort` command to the Keys api
sort is the only command that's missing from the list. 

refer [here](https://redis.io/commands/sort)

sloanesturz:
Fixes #218.

SORT has two different response types:

```
SORT myKey # -> returns the sorted values in the list stored at myKey
SORT myKey STORE resultKey # -> returns the number of elements sorted, writes values to resultKey
```

This is not easily expressible in Scala, so I created two separate functions, one for `SORT ... STORE ...` and one for `SORT ...`

This is my first commit to zio-redis, please let me know if I am not following any important functional or stylistic guidelines!

### DIFF
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -138,11 +138,6 @@ object Input {
       Chunk(encodeString("LIMIT"), encodeString(data.offset.toString), encodeString(data.count.toString))
   }
 
-  final case class ListInput[-A](input: Input[A]) extends Input[List[A]] {
-    override private[redis] def encode(data: List[A]): Chunk[RespValue.BulkString] =
-      Chunk.fromIterable(data).flatMap(input.encode)
-  }
-
   case object LongInput extends Input[Long] {
     def encode(data: Long): Chunk[RespValue.BulkString] = Chunk.single(encodeString(data.toString))
   }
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -34,6 +34,14 @@ object Input {
     def encode(data: Boolean): Chunk[RespValue.BulkString] = Chunk.single(encodeString(if (data) "1" else "0"))
   }
 
+  case object ByInput extends Input[String] {
+    def encode(data: String): Chunk[RespValue.BulkString] = Chunk(encodeString("BY"), encodeString(data))
+  }
+
+  case object GetInput extends Input[String] {
+    def encode(data: String): Chunk[RespValue.BulkString] = Chunk(encodeString("GET"), encodeString(data))
+  }
+
   case object BitFieldCommandInput extends Input[BitFieldCommand] {
     def encode(data: BitFieldCommand): Chunk[RespValue.BulkString] = {
       import BitFieldCommand._
@@ -144,6 +152,11 @@ object Input {
     def encode(data: Unit): Chunk[RespValue.BulkString] = Chunk.empty
   }
 
+  final case class ListInput[-A](input: Input[A]) extends Input[List[A]] {
+    override private[redis] def encode(data: List[A]): Chunk[RespValue.BulkString] =
+      data.foldLeft(Chunk.empty: Chunk[RespValue.BulkString])((acc, a) => acc ++ input.encode(a))
+  }
+
   final case class NonEmptyList[-A](input: Input[A]) extends Input[(A, List[A])] {
     def encode(data: (A, List[A])): Chunk[RespValue.BulkString] =
       (data._1 :: data._2).foldLeft(Chunk.empty: Chunk[RespValue.BulkString])((acc, a) => acc ++ input.encode(a))
@@ -151,6 +164,10 @@ object Input {
 
   case object OrderInput extends Input[Order] {
     def encode(data: Order): Chunk[RespValue.BulkString] = Chunk.single(encodeString(data.stringify))
+  }
+
+  case object AlphaInput extends Input[Alpha] {
+    def encode(data: Alpha): Chunk[RespValue.BulkString] = Chunk.single(encodeString(data.stringify))
   }
 
   case object RadiusUnitInput extends Input[RadiusUnit] {
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -26,20 +26,16 @@ object Input {
       Chunk(encodeString("AGGREGATE"), encodeString(data.stringify))
   }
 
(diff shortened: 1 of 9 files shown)


redis/src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -216,6 +216,57 @@ trait Keys {
...
+   * @param limitOffset limit count and offset option, take only limit values, starting at position offset
+   * @param desc desc option, sort descending instead of ascending
+   * @param get get option, return the values referenced by the keys generated from the get patterns
+   * @param alpha alpha option, sort the values alphanumerically, instead of by interpreting the value as floating point number
+   * @return the sorted values, or the values found using the get patterns
+   */
+  final def sort(
+    key: String,
+    by: Option[String] = None,
+    limitOffset: Option[(Int, Int)] = None,
```
mijicd, line 233: We have this defined among sorted sets (see [here](https://github.com/zio/zio-redis/blob/b72e4fc4b98493baf5ae782868517cbedfdd9369/redis/src/main/scala/zio/redis/options/SortedSets.scala#L63)). I propose that you reuse it and move the option itself to shared options.

redis/src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -216,6 +216,57 @@ trait Keys {
...
+    desc: Boolean = false,
```
mijicd, line 234: Order is defined as geo options [here](https://github.com/zio/zio-redis/blob/b72e4fc4b98493baf5ae782868517cbedfdd9369/redis/src/main/scala/zio/redis/options/Geo.scala#L11). Please do the same approach:
- move to shared
- use it instead of plain boolean

mijicd (commented):
@sloanesturz thank you for taking care of this! I left some general remarks, but more specifically, I don't think we should expose two separate commands. I would propose to:
- apply the proposed fixes
- represent missing terms (e.g. alpha)
- unify commands into single sort

Please take a look at geo API, as there's quite a few options defined there, and it might serve as an "inspiration" here :).

sloanesturz:
OK, thanks for the pointers to those options fields -- definitely more clean to use those instead of my approach. The pattern in the Geo package has a lot of overlap and is clearly the right way to go.

> unify commands into single sort

What should the return type of this method be? If we `SORT ... STORE ...`, Redis will return a number. If we `SORT ...`, Redis will return a list of values. To unify these two I could always return a `Chunk[String]` and if you had happened to `SORT ... STORE ...` you just "have to know" that the Chunk[String] will have one entry, the number of values. To me, that seems pretty kludgy but I suppose it does sorta replicate what Redis is doing.

mijicd:
One option might be to define a custom response type (ADT of possible responses), but let's play around with this, perhaps something better pops up.

sloanesturz:
OK, let me take a crack at your first suggestions before taking on the second one. In my experience, returning an `Either[Long, Seq[String]]` that is fully is defined by the kinds of arguments to the function is probably not what we want here because you end up writing code like

```
redis.sort(key, storeAt = None).map {
  case _: Right => ??? // as a developer, I know this can't happen, annoying the compiler makes me handle this
  case _: Left =>  doMyImplementation() // the actual implementation I want to do
}
```

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -126,6 +138,11 @@ object Input {
       Chunk(encodeString("LIMIT"), encodeString(data.offset.toString), encodeString(data.count.toString))
   }
 
+  final case class ListInput[-A](input: Input[A]) extends Input[List[A]] {
+    override private[redis] def encode(data: List[A]): Chunk[RespValue.BulkString] =
+      data.foldLeft(Chunk.empty: Chunk[RespValue.BulkString])((acc, a) => acc ++ input.encode(a))
```
sloanesturz, line 143: compiler was very unhappy about `Chunk(data.map(input.encode))` ... may come back to figure out how to make this more clear.

redis/src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -266,6 +313,20 @@ trait Keys {
...
+    get: List[String],
+    alpha: Boolean
+  ): List[String] =
+    by.map(sortBy => List("BY", sortBy)).getOrElse(List.empty) ++
+      limitOffset.map { case (count, offset) => List("LIMIT", offset.toString, count.toString) }
+        .getOrElse(List.empty) ++
+      get.flatMap(getAt => List("GET", getAt)) ++
+      List(if (desc) "DESC" else "ASC") ++
+      (if (alpha) List("ALPHA") else List.empty)
+
```
mijicd, line 329: Should this be deleted?

sloanesturz: Fixed!

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -126,6 +138,11 @@ object Input {
       Chunk(encodeString("LIMIT"), encodeString(data.offset.toString), encodeString(data.count.toString))
   }
 
+  final case class ListInput[-A](input: Input[A]) extends Input[List[A]] {
+    override private[redis] def encode(data: List[A]): Chunk[RespValue.BulkString] =
```
mijicd, line 142:
```suggestion
    def encode(data: List[A]): Chunk[RespValue.BulkString] =
```

However, see the comment below, this might be unnecessary.

redis/src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -346,6 +393,37 @@ private[redis] object Keys {
...
 
+  final val Sort
+    : RedisCommand[(String, Option[String], Option[Limit], List[String], Order, Option[Alpha]), Chunk[String]] =
+    RedisCommand(
+      "SORT",
+      Tuple6(
+        StringInput,
+        OptionalInput(ByInput),
+        OptionalInput(LimitInput),
+        ListInput(GetInput),
```
mijicd, line 404: This can actually be changed to `OptionalInput(NonEmptyList[GetInput])`.

mijicd (commented):
Almost there! Since we opted for this approach of command per output (with a good reason, as you elaborated), do you mind doing a quick scan of the similar commands and filing them in an issue so that we can align the API? :pray:

redis/src/main/scala/zio/redis/api/Keys.scala
sloanesturz: clever!

sloanesturz:
Filed #306 for the GEO functions.

PR #303, opened 2021-03-05
add testExecutor for hyperLogLog

jxnu-liguobin:
Closes https://github.com/zio/zio-redis/issues/193

### DIFF
--- a/redis/src/main/scala/zio/redis/TestExecutor.scala
+++ b/redis/src/main/scala/zio/redis/TestExecutor.scala
@@ -10,7 +10,8 @@ import zio.stm.{ STM, TMap, TRef, USTM }
 private[redis] final class TestExecutor private (
   sets: TMap[String, Set[String]],
   strings: TMap[String, String],
-  randomPick: Int => USTM[Int]
+  randomPick: Int => USTM[Int],
+  hyperLogLogs: TMap[String, Seq[String]]
 ) extends RedisExecutor.Service {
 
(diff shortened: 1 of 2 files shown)


redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -257,7 +260,48 @@ private[redis] final class TestExecutor private (
...
+            }
+            .map(vs => RespValue.Integer(vs.size.toLong)),
+          STM.succeedNow(Replies.WrongType)
+        )
+      case api.HyperLogLog.PfMerge.name =>
+        val key    = input.head.asString
+        val values = input.tail.map(_.asString)
+        STM.ifM(forAll(values ++ Chunk.single(key))(isSet))(
+          for {
+            sourceValues <- STM.foldLeft(values)(Seq.empty[String]) { (bHyperLogLogs, nextKey) =>
```
mijicd, line 294: I'd use chunk instead of a Seq.

mijicd (commented):
Just one super small nit, other than that looks good to go.

redis/src/main/scala/zio/redis/TestExecutor.scala
jxnu-liguobin: Chunk.empty  generice type is Nothing,can not use ++ with Chunk String

mijicd (approved):

PR #314, opened 2021-03-14
API for binary data

### LINKED ISSUE #214: API for binary data

anatolysergeev:
Closes #214


redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -2,44 +2,46 @@ package zio.redis
...
-  private[redis] def encode(data: A): Chunk[RespValue.BulkString]
+  private[redis] def encode(data: A)(implicit codec: Codec): Chunk[RespValue.BulkString]
 }
 
 object Input {
 
   @inline
   private[this] def encodeString(s: String): RespValue.BulkString = RespValue.bulkString(s)
+  @inline
+  private[this] def encodeBytes(s: Chunk[Byte]): RespValue.BulkString = RespValue.BulkString(s)
```
mijicd, line 20: I presume `encodeString` we'll eventually be completely removed.

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -2,44 +2,46 @@ package zio.redis
...
 import scala.util.matching.Regex
-
 import zio.Chunk
 import zio.duration.Duration
+import zio.schema.codec.Codec
+import zio.schema.Schema
 
 sealed trait Input[-A] {
-  private[redis] def encode(data: A): Chunk[RespValue.BulkString]
+  private[redis] def encode(data: A)(implicit codec: Codec): Chunk[RespValue.BulkString]
```
mijicd, line 12: I wouldn't pass it implicitly, please change to the encoding suggested in our chat:

```scala
def encode(data: A, codec: Codec): Chunk[...]
```

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -426,42 +435,74 @@ object Input {
...
+  case object EvalInput extends Input[(String, Chunk[Chunk[Byte]], Chunk[Chunk[Byte]])] {
+    def encode(data: (String, Chunk[Chunk[Byte]], Chunk[Chunk[Byte]]))(implicit codec: Codec): Chunk[RespValue.BulkString] = {
+      val (lua, keys, args) = data
+      val encodedScript     = Chunk(encodeString(lua), encodeString(keys.size.toString))
+      val encodedKeys       = keys.foldLeft[Chunk[RespValue.BulkString]](Chunk.empty)((cur, next) => cur :+ encodeBytes(next))
+      val encodedArgs       = args.foldLeft[Chunk[RespValue.BulkString]](Chunk.empty)((cur, next) => cur :+ encodeBytes(next))
+      encodedScript ++ encodedKeys ++ encodedArgs
+    }
+  }
+
```
mijicd, line 462: Scripting part should stay in its own pull request, please don't introduce it here.

redis/src/main/scala/zio/redis/RedisCommand.scala
```diff
@@ -4,10 +4,7 @@ import zio.ZIO
 
 final class RedisCommand[-In, +Out] private (val name: String, val input: Input[In], val output: Output[Out]) {
   private[redis] def run(in: In): ZIO[RedisExecutor, RedisError, Out] =
-    ZIO
-      .accessM[RedisExecutor](_.get.execute(Input.StringInput.encode(name) ++ input.encode(in)))
-      .flatMap(out => ZIO.effect(output.unsafeDecode(out)))
-      .refineToOrDie[RedisError]
+    ZIO.accessM[RedisExecutor](_.get.execute(name, in)(input, output))
```
mijicd, line 7: I wouldn't change this. Instead, I'd just expose codec in executor.

redis/src/main/scala/zio/redis/RedisExecutor.scala
```diff
@@ -2,17 +2,21 @@ package zio.redis
...
 import zio._
 import zio.logging._
+import zio.schema.codec.Codec
 
 object RedisExecutor {
   trait Service {
+
     def execute(command: Chunk[RespValue.BulkString]): IO[RedisError, RespValue]
+
+    def execute[In: Input, Out: Output](name: String, in: In): IO[RedisError, Out]
```
mijicd, line 12: As suggested in the previous comment, let's drop this method and replace it with:

```scala
def codec: Codec
```

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -14,19 +16,31 @@ private[redis] final class TestExecutor private (
...
   def execute(command: Chunk[RespValue.BulkString]): IO[RedisError, RespValue] =
     for {
       name   <- ZIO.fromOption(command.headOption).orElseFail(ProtocolError("Malformed command."))
       result <- runCommand(name.asString, command.tail).commit
     } yield result
 
+  def execute[In: Input, Out: Output](name: String, in: In): IO[RedisError, Out] = {
+    val zio = for {
+      respOut <- runCommand(name, implicitly[Input[In]].encode(in)(codec)).commit
+      decodedOut <- ZIO.effect(implicitly[Output[Out]].unsafeDecode(respOut))
```
mijicd, line 30: As suggested in the previous comment this method should be removed. However, hint for future: whenever you end up invoking `implicitly[something]` like here, consider adding the so called summoner to the typeclass (if you can, as here). For example:

```scala
object Input {
  def apply[A](implicit instance: Input[A]) = instance
}
```

That way, the call site becomes just:

```scala
Input[In].encode(...)
```

mijicd (commented):
Going in the right direction!

redis/src/main/scala/zio/redis/Input.scala
anatolysergeev: Ok)
I want to mention that if it is not implicit parameter, then we have to put codec parameter in every encode method inside that one
`final case class Tuple5[-A, -B, -C, -D, -E](_1: Input[A], _2: Input[B], _3: Input[C], _4: Input[D], _5: Input[E])
      extends Input[(A, B, C, D, E)] {
    def encode(data: (A, B, C, D, E))(implicit codec: Codec): Chunk[RespValue.BulkString] =
      _1.encode(data._1) ++ _2.encode(data._2) ++ _3.encode(data._3) ++ _4.encode(data._4) ++ _5.encode(data._5)
  }`

mijicd: I see, let's keep it then. We'll deal with compat later.

redis/src/main/scala/zio/redis/codec/StringUtf8Codec.scala
```diff
@@ -0,0 +1,47 @@
...
+import java.nio.charset.StandardCharsets
+
+import zio.redis.RedisError.CodecError
+import zio.schema.Schema
+import zio.schema.StandardType.StringType
+import zio.schema.codec.Codec
+import zio.stream.ZTransducer
+import zio.{ Chunk, ZIO }
+
+object StringUtf8Codec extends Codec {
```
anatolysergeev, line 12: That codec is very poor, i've been considering to do primitive (short, long, etc.) codec 🤔  But we had agreement to do codec only for strings and i decided to stick with that

anatolysergeev:
Add changes only for sets commands, that PR already very huge(((

redis/src/main/scala/zio/redis/RedisCommand.scala
```diff
@@ -5,8 +5,12 @@ import zio.ZIO
 final class RedisCommand[-In, +Out] private (val name: String, val input: Input[In], val output: Output[Out]) {
   private[redis] def run(in: In): ZIO[RedisExecutor, RedisError, Out] =
     ZIO
-      .accessM[RedisExecutor](_.get.execute(Input.StringInput.encode(name) ++ input.encode(in)))
-      .flatMap(out => ZIO.effect(output.unsafeDecode(out)))
+      .accessM[RedisExecutor] { executor =>
+        val service = executor.get
+        val codec   = service.codec
+        val command = Input.StringInput.encode(name)(codec) ++ input.encode(in)(codec)
+        service.execute(command).flatMap[Any, Throwable, Out](out => ZIO.effect(output.unsafeDecode(out)(codec)))
```
anatolysergeev, line 12: Interesting thing without explicit types in flatMap scala 2.12.13 doesn't compile with error `a type was inferred to be `Any`; this may indicate a programming error.`

mijicd: Yes, this is a very annoying issue.

benchmarks/src/main/scala/zio/redis/BenchmarksUtils.scala
```diff
@@ -16,5 +18,6 @@ trait BenchmarksUtils {
 }
 
 object BenchmarksUtils {
-  private final val Layer = Logging.ignore >>> RedisExecutor.local.orDie
+  private final val Codec: ULayer[Has[Codec]] = ZLayer.succeed(StringUtf8Codec)
```
mijicd, line 21: I'd "inline" this into `Layer` definition if possible.

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -54,192 +62,220 @@ object Input {
...
+    def encode(data: ScoreRange)(implicit codec: Codec): Chunk[RespValue.BulkString] =
       Chunk(encodeString(data.min.stringify), encodeString(data.max.stringify))
   }
 
   case object StringInput extends Input[String] {
-    def encode(data: String): Chunk[RespValue.BulkString] = Chunk.single(encodeString(data))
+    def encode(data: String)(implicit codec: Codec): Chunk[RespValue.BulkString] = Chunk.single(encodeString(data))
+  }
+
+  final case class ArbitraryInput[A: Schema]() extends Input[A] {
```
mijicd, line 228: You can even make it an "explicit implicit", e.g.

```scala
final case class ArbitraryInput[A](implicit schema: Schema[A]) extends Input[A] { ... }
```

That way you don't have to use the summoner.

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -122,56 +114,58 @@ object Output {
...
       respValue match {
         case RespValue.BulkString(value) => value
         case other                       => throw ProtocolError(s"$other isn't a bulk string")
       }
   }
 
+  final case class ArbitraryOutput[A: Schema]() extends Output[A] {
+    protected def tryDecode(respValue: RespValue)(implicit codec: Codec): A =
+      respValue match {
+        case RespValue.BulkString(s) => codec.decode(Schema[A])(s).fold(e => throw CodecError(e), identity)
```
mijicd, line 143: `getOrElse` is probably nicer than `fold` in this situation.

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -290,15 +284,15 @@ object Output {
...
+    protected def tryDecode(respValue: RespValue)(implicit codec: Codec): Map[String, Map[String, String]] =
       respValue match {
         case RespValue.NullArray       => Map.empty[String, Map[String, String]]
         case RespValue.Array(entities) => extractMapMap(entities)
         case other                     => throw ProtocolError(s"$other isn't an array")
       }
   }
 
-  private def extractMapMap(entities: Chunk[RespValue]): Map[String, Map[String, String]] = {
+  private def extractMapMap(entities: Chunk[RespValue])(implicit codec: Codec): Map[String, Map[String, String]] = {
```
mijicd, line 295: Non-related, but let's rename to `extractNestedMap` :).

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -27,6 +32,22 @@ private[redis] final class TestExecutor private (
...
+  private val SIsMemberCommand   = api.Sets.SIsMember[String, String]
+  private val SMembersCommand    = api.Sets.SMembers[String, String]
+  private val SMoveCommand       = api.Sets.SMove[String, String, String]
+  private val SPopCommand        = api.Sets.SPop[String, String]
+  private val SRandMemberCommand = api.Sets.SRandMember[String, String]
+  private val SRemCommand        = api.Sets.SRem[String, String]
+  private val SUnionCommand      = api.Sets.SUnion[String, String]
+  private val SUnionStoreCommand = api.Sets.SUnionStore[String, String]
+  private val SScanCommand       = api.Sets.SScan[String, String]
+
```
mijicd, line 50: I'd leave this one as it was before, just to reduce the diff, as we'll have to address the test executor thoroughly anyhow. If that's not possible, please ignore this comment.

redis/src/main/scala/zio/redis/api/Sets.scala
```diff
@@ -172,59 +179,66 @@ trait Sets {
...
+    SUnionStore[D, K].run((destination, (key, keys.toList)))
 }
 
 private[redis] object Sets {
-  final val SAdd: RedisCommand[(String, (String, List[String])), Long] =
-    RedisCommand("SADD", Tuple2(StringInput, NonEmptyList(StringInput)), LongOutput)
 
-  final val SCard: RedisCommand[String, Long] = RedisCommand("SCARD", StringInput, LongOutput)
+  final def SAdd[K: Schema, M: Schema]: RedisCommand[(K, (M, List[M])), Long] =
+    RedisCommand("SADD", Tuple2(ArbitraryInput(), NonEmptyList(ArbitraryInput())), LongOutput)
```
mijicd, line 189: They're not constants anymore, so it might make sense to just inline them to the client calls above, e.g.

```scala
final def sAdd[K: Schema, M: Schema](key: K, member: M, members: M*): ZIO[RedisExecutor, RedisError, Long] = {
  val command = RedisCommand("SADD", Tuple2(ArbitraryInput(), NonEmptyList(ArbitraryInput())), LongOutput)
  
  command.run((key, (member, members.toList)))
}
```

I'm fine if you decide not to do it now to minimize the diff though.

redis/src/main/scala/zio/redis/codec/StringUtf8Codec.scala
mijicd: It's fine for now, we can improve once the API is stabilized (or hopefully delete, once zio-schema supports it).

mijicd (commented):
Going in the right direction :rocket:

redis/src/main/scala/zio/redis/api/Sets.scala
anatolysergeev: Done

redis/src/main/scala/zio/redis/TestExecutor.scala
anatolysergeev: Yeah, that's brings compilation errors(
But I pulled out command names from RedisCommand into constants, and has returned almost like was before, without names. What do u think about that?

redis/src/main/scala/zio/redis/Output.scala
anatolysergeev: Done

redis/src/main/scala/zio/redis/Output.scala
anatolysergeev: Yeah, but the thing is that getOrElse doesn't give us a value in Left(v) :( I wanted to throw a specific error.

redis/src/main/scala/zio/redis/Input.scala
anatolysergeev: Done

benchmarks/src/main/scala/zio/redis/BenchmarksUtils.scala
anatolysergeev: Done

build.sbt
```diff
@@ -37,10 +37,11 @@ lazy val redis =
     .settings(buildInfoSettings("zio.redis"))
     .settings(
       libraryDependencies ++= Seq(
-        "dev.zio" %% "zio-streams"  % Zio,
-        "dev.zio" %% "zio-logging"  % "0.5.8",
-        "dev.zio" %% "zio-test"     % Zio % Test,
-        "dev.zio" %% "zio-test-sbt" % Zio % Test
+        "dev.zio" %% "zio-streams"     % Zio,
+        "dev.zio" %% "zio-logging"     % "0.5.8",
+        "dev.zio" %% "zio-schema-core" % "0.0.1+5-be72f2ee-SNAPSHOT",
```
mijicd, line 42: Please drop the snapshot dependency and use `0.0.4`.

redis/src/main/scala/zio/redis/RedisExecutor.scala
```diff
@@ -23,24 +26,26 @@ object RedisExecutor {
...
     reqQueue: Queue[Request],
     resQueue: Queue[Promise[RedisError, RespValue]],
     byteStream: ByteStream.Service,
-    logger: Logger[String]
+    logger: Logger[String],
+    override val codec: Codec
   ) extends Service {
 
-    def execute(command: Chunk[RespValue.BulkString]): IO[RedisError, RespValue] =
+    override def execute(command: Chunk[RespValue.BulkString]): IO[RedisError, RespValue] =
```
mijicd, line 48:
```suggestion
    def execute(command: Chunk[RespValue.BulkString]): IO[RedisError, RespValue] =
```

mijicd (commented):
Just two small comments, looks good to go otherwise.

project/BuildHelper.scala
```diff
@@ -29,6 +29,7 @@ object BuildHelper {
         "com.github.liancheng" %% "organize-imports" % "0.5.0",
         "com.github.vovapolu"  %% "scaluzzi"         % "0.1.18"
       ),
+      resolvers += "Sonatype OSS Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots",
```
mijicd, line 32:
```suggestion
```

mijicd (approved):

mijicd:
Amazing work :rocket: :tada:

Issue #320, opened 2021-03-18
Scan command results in error when used with Count and or Type option


hcwilhelm:
When using the Scan command with Count and or Type option it results in `ProtocolError: syntax error` because of incorrect encodings for the input types `OptionalInput(LongInput)` and `OptionalInput(StringInput)`.

Also although the encoding for the `RegexInput` is correct the use of the `Regex` type itself doesn't fit the `glob-style patterns` which are used by Redis https://redis.io/commands/keys. So to use these `glob-style` patterns the underlying type for the `RegexInput` should better be a `String` instead of a `Regex`.

hcwilhelm:
I take this one.

PR #321, opened 2021-03-18
Fix scan commands

hcwilhelm:
Closes: #320 

Fix all scan functions scan, sscan, hscan and zscan to properly support `Pattern`, `Count` and `Type` options.

### DIFF
--- a/redis/src/main/scala/zio/redis/api/Keys.scala
+++ b/redis/src/main/scala/zio/redis/api/Keys.scala
@@ -6,7 +6,7 @@ import zio.duration._
 import zio.redis.Input._
 import zio.redis.Output._
 import zio.redis._
-import zio.{ Chunk, Task, ZIO }
+import zio.{ Chunk, ZIO }
 
 trait Keys {
   import Keys.{ Keys => _, _ }
--- a/redis/src/main/scala/zio/redis/api/Keys.scala
+++ b/redis/src/main/scala/zio/redis/api/Keys.scala
@@ -2,13 +2,11 @@ package zio.redis.api
 
 import java.time.Instant
 
-import scala.util.matching.Regex
-
 import zio.duration._
 import zio.redis.Input._
 import zio.redis.Output._
 import zio.redis._
-import zio.{ Chunk, ZIO }
+import zio.{ Chunk, Task, ZIO }
 
 trait Keys {
   import Keys.{ Keys => _, _ }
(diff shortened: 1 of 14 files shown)


redis/src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -211,10 +209,13 @@ trait Keys {
    */
   final def scan(
     cursor: Long,
-    pattern: Option[Regex] = None,
+    pattern: Option[String] = None,
     count: Option[Long] = None,
     `type`: Option[String] = None
-  ): ZIO[RedisExecutor, RedisError, (Long, Chunk[String])] = Scan.run((cursor, pattern, count, `type`))
+  ): ZIO[RedisExecutor, RedisError, (Long, Chunk[String])] =
+    Task.effect(`type`.map(RedisType.from)).refineOrDie { case e: RedisError => e } >>= (redisType =>
```
mijicd, line 216: Let's accept RedisType instead and remove this parsing step completely.

redis/src/main/scala/zio/redis/api/Sets.scala
```diff
@@ -143,16 +141,17 @@ trait Sets {
...
    * @return Returns the next cursor, and items for this iteration or nothing when you reach the end, as a tuple
    */
   final def sScan(
     key: String,
     cursor: Long,
-    regex: Option[Regex] = None,
-    count: Option[Count] = None
-  ): ZIO[RedisExecutor, RedisError, (Long, Chunk[String])] = SScan.run((key, cursor, regex, count))
+    pattern: Option[String] = None,
+    count: Option[Long] = None
```
mijicd, line 152: I think it's fine to have `Count` here (propagate to all others as it's not in sync).

redis/src/test/scala/zio/redis/KeysSpec.scala
```diff
@@ -92,7 +92,51 @@ trait KeysSpec extends BaseSpec {
...
+        } yield assert(scan)(fails(isSubtype[RedisError.WrongType](anything)))
+      },
+      testM("scan entries with match, count and type options") {
+        for {
+          key             <- uuid
+          value           <- uuid
+          _               <- set(key, value)
+          scan            <- scan(0L, pattern = Some("*"), count = Some(100), `type` = Some("zset"))
+          (next, elements) = scan
+        } yield assert(next)(isGreaterThanEqualTo(0L)) && assert(elements)(isNonEmpty)
```
mijicd, line 139: Not saying that you should do it , but you could probably replace a lot of duplicates by rewriting scan tests as property tests, as it would explore the argument combinations in a compact way.

mijicd (commented):
Looks good, just a few tweaks. Also, it seems to me that the root problem was actually poor encoding of type, I still haven't quite understand why count would make any problems as it was encoded properly (unlike the type).

hcwilhelm:
> Looks good, just a few tweaks. Also, it seems to me that the root problem was actually poor encoding of type, I still haven't quite understand why count would make any problems as it was encoded properly (unlike the type).

Yea count already had a proper encoding, but it wasn't used everywhere and especially in the Keys#scan function it wasn't used.

redis/src/main/scala/zio/redis/api/Keys.scala
hcwilhelm: Change to RedisType

redis/src/main/scala/zio/redis/api/Sets.scala
hcwilhelm: Changed back to Count

redis/src/test/scala/zio/redis/KeysSpec.scala
hcwilhelm: Used property based testing for this.

mijicd (dismissed):
Nice work!

mijicd (approved):

PR #322, opened 2021-03-19
Add LCS functionality

### LINKED ISSUE #155: Implement remaining string commands

### Commands

- [ ] STRALGO LCS algo-specific-argument [algo-specific-argument ...] Run algorithms (currently LCS) against strings

### Tips

barthorre:
Closes #155


redis/src/main/scala/zio/redis/options/Strings.scala
```diff
@@ -2,6 +2,25 @@ package zio.redis.options
...
+
+  object StralgoCommand {
+    sealed case class StralgoLCS(`type`: StralgoType) extends StralgoCommand
+  }
+
+  sealed trait StralgoType { self =>
+    private[redis] final def stringify: String =
+      self match {
+        case StralgoType.Strings => "STRINGS"
+        case StralgoType.Keys    => "KEYS"
```
jxnu-liguobin, line 15: I'm just looking around. Some options don't seem to be supported. List them? `[LEN] [IDX] [MINMATCHLEN <len>] [WITHMATCHLEN]
`

barthorre: true, some of these would also alter the output, that's why I'd like some feedback on the input command structure before delving deeper into the other options.

redis/src/main/scala/zio/redis/options/Strings.scala
```diff
@@ -2,6 +2,25 @@ package zio.redis.options
 
 trait Strings {
 
+  sealed trait StralgoCommand
+
+  object StralgoCommand {
+    sealed case class StralgoLCS(`type`: StralgoType) extends StralgoCommand
+  }
```
mijicd, line 9: I'd avoid the command suffix as it's heavily used for redis commands. Additionally, perhaps it makes sense to unify the two, e.g.:

```scala
sealed trait Algorithm

object Algorithm {
  case object StringLcs extends Algorithm
  case object KeysLcs extends Algorithm
}
```

Of course, you can (and should) further expand with parameters as discussed by you and @jxnu-liguobin .

mijicd (commented):
It goes in the right direction.

redis/src/main/scala/zio/redis/options/Strings.scala
jxnu-liguobin: it looks really good to use the unsuffixed one. It is more easy to illustrate the relationship of subcommands

redis/src/main/scala/zio/redis/options/Strings.scala
```diff
@@ -2,6 +2,38 @@ package zio.redis.options
...
+  object StralgoLCS {
+    case object Strings extends StrAlgoLCS
+    case object Keys    extends StrAlgoLCS
+  }
+
+  sealed trait StrAlgoLcsQueryType
+
+  object StrAlgoLcsQueryType {
+    case object Len                                                           extends StrAlgoLcsQueryType
+    case class Idx(minMatchLength: Int = 1, withMatchLength: Boolean = false) extends StrAlgoLcsQueryType
```
anovakovic01, line 23: It's not so clear from the documentation, but it looks like `MINMATCHLEN` can be specified without `IDX`.

barthorre: it indeed seems like you can, but it doesn;t seem to do anything

image

anovakovic01: I agree, so let's keep it where it already is.

redis/src/main/scala/zio/redis/options/Strings.scala
barthorre: I opted for the unsuffixed ones as that, to me, reflects the documentation/stucture of redis commands better.

anovakovic01 (approved):

mijicd (approved):

Issue #312, opened 2021-03-13
Add the FULL option to XINFO STREAM command


anovakovic01:
There is no support for FULL option implemented in `zio-redis`. This is not a priority as it works only with Redis version 6.

jxnu-liguobin:
doing. 

I found that some mainstream clients do not support this option. Because the return structure of this option is extremely complex, I may need a new command.(It's very painful to parse the structure)

jxnu-liguobin:
closed by #323

PR #323, opened 2021-03-22
Add full to xinfo stream

jxnu-liguobin:
#312

The internal structure naming references to [redis/redis](https://github.com/redis/redis/blob/6da2bc96e210dba081328d9b24f466cea6047826/src/t_stream.c#L3332), but external naming such as `FullStreamInfo`,`xInfoStreamFull ` may need to be modified and need more testing.

### DIFF
--- a/redis/src/main/scala/zio/redis/Output.scala
+++ b/redis/src/main/scala/zio/redis/Output.scala
@@ -1,7 +1,5 @@
 package zio.redis
 
-import scala.collection.mutable.ListBuffer
-
 import zio.Chunk
 import zio.duration._
 
@@ -327,7 +325,7 @@ object Output {
                   case (key @ RespValue.BulkString(_), value @ RespValue.BulkString(_)) =>
                     if (key.asString == XInfoFields.Name)
                       streamGroupsInfo = streamGroupsInfo.copy(name = value.asString)
-                    else if (key.asString == XInfoFields.LastDelivered)
+                    else if (key.asString == XInfoFields.LastDeliveredId)
                       streamGroupsInfo = streamGroupsInfo.copy(lastDeliveredId = value.asString)
                   case (key @ RespValue.BulkString(_), value @ RespValue.Integer(_)) =>
                     if (key.asString == XInfoFields.Pending)
@@ -387,9 +385,9 @@ object Output {
       }
   }
 
-  case object StreamInfoFullOutput extends Output[XInfoFullStream.StreamFullInfo] {
-    override protected def tryDecode(respValue: RespValue): XInfoFullStream.StreamFullInfo = {
-      var streamInfoFull: XInfoFullStream.StreamFullInfo = XInfoFullStream.StreamFullInfo.empty
+  case object StreamInfoFullOutput extends Output[StreamInfoWithFull.FullStreamInfo] {
+    override protected def tryDecode(respValue: RespValue): StreamInfoWithFull.FullStreamInfo = {
+      var streamInfoFull: StreamInfoWithFull.FullStreamInfo = StreamInfoWithFull.FullStreamInfo.empty
       respValue match {
         // Note that you should not rely on the fields exact position. see https://redis.io/commands/xinfo
         case RespValue.Array(elements) if elements.length % 2 == 0 =>
@@ -406,11 +404,12 @@ object Output {
               case (key @ RespValue.BulkString(_), value @ RespValue.BulkString(_))
                   if key.asString == XInfoFields.LastGeneratedId =>
                 streamInfoFull = streamInfoFull.copy(lastGeneratedId = value.asString)
-              case (key @ RespValue.BulkString(_), value @ RespValue.Array(_)) if key.asString == XInfoFields.Entries =>
-                streamInfoFull = streamInfoFull.copy(entries = Some(extractStreamEntry(value)))
+              case (key @ RespValue.BulkString(_), RespValue.Array(values)) if key.asString == XInfoFields.Entries =>
+                val streamEntryList = values.map(extractStreamEntry)
+                streamInfoFull = streamInfoFull.copy(entries = streamEntryList)
               case (key @ RespValue.BulkString(_), RespValue.Array(value)) if key.asString == XInfoFields.Groups =>
                 // Get the group list of the stream.
-                streamInfoFull = streamInfoFull.copy(groups = extractXInfoFullGroups(value).toSeq)
+                streamInfoFull = streamInfoFull.copy(groups = extractXInfoFullGroups(value))
               case _ =>
             }
             pos += 2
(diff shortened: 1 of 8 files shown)


redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -385,6 +385,152 @@ object Output {
...
+          var pos = 0
+          while (pos < elements.length) {
+            (elements(pos), elements(pos + 1)) match {
+              // Get the basic information of the outermost stream.
+              case (key @ RespValue.BulkString(_), value @ RespValue.Integer(_)) =>
+                if (key.asString == XInfoFields.Length) streamInfoFull = streamInfoFull.copy(length = value.value)
+                else if (key.asString == XInfoFields.RadixTreeNodes)
+                  streamInfoFull = streamInfoFull.copy(radixTreeNodes = value.value)
+                else if (key.asString == XInfoFields.RadixTreeKeys)
+                  streamInfoFull = streamInfoFull.copy(radixTreeKeys = value.value)
```
anovakovic01, line 403: Think about using `match case` here.

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -385,6 +385,152 @@ object Output {
...
+              case (key @ RespValue.BulkString(_), value @ RespValue.BulkString(_))
+                  if key.asString == XInfoFields.LastGeneratedId =>
+                streamInfoFull = streamInfoFull.copy(lastGeneratedId = value.asString)
+              case (key @ RespValue.BulkString(_), RespValue.Array(values)) if key.asString == XInfoFields.Entries =>
+                val streamEntryList = values.map(extractStreamEntry)
+                streamInfoFull = streamInfoFull.copy(entries = streamEntryList)
+              case (key @ RespValue.BulkString(_), RespValue.Array(value)) if key.asString == XInfoFields.Groups =>
+                // Get the group list of the stream.
+                streamInfoFull = streamInfoFull.copy(groups = extractXInfoFullGroups(value))
```
anovakovic01, line 412: I suggest that you create the `extractXInfoFullGroup` function and use `values.map` here, just like you did for an `extractStreamEntry`.

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -385,6 +385,152 @@ object Output {
...
+                if (key.asString == XInfoFields.LastDeliveredId)
+                  readyGroup = readyGroup.copy(lastDeliveredId = value.asString)
+                else if (key.asString == XInfoFields.Name) readyGroup = readyGroup.copy(name = value.asString)
+              case (key @ RespValue.BulkString(_), value @ RespValue.Integer(_))
+                  if XInfoFields.PelCount == key.asString =>
+                readyGroup = readyGroup.copy(pelCount = value.value)
+              case (key @ RespValue.BulkString(_), RespValue.Array(value)) =>
+                // Get the consumer list of the current group.
+                if (key.asString == XInfoFields.Consumers) {
+                  readyGroup = readyGroup.copy(consumers = extractXInfoFullConsumers(value))
```
anovakovic01, line 446: I suggest that you create the `extractXInfoFullConsumer` function and use `values.map` here, just like you did for an `extractStreamEntry`.

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -385,6 +385,152 @@ object Output {
...
+            (groupElements(groupElementPos), groupElements(groupElementPos + 1)) match {
+              // Get the basic information of the current group.
+              case (key @ RespValue.BulkString(_), value @ RespValue.BulkString(_)) =>
+                if (key.asString == XInfoFields.LastDeliveredId)
+                  readyGroup = readyGroup.copy(lastDeliveredId = value.asString)
+                else if (key.asString == XInfoFields.Name) readyGroup = readyGroup.copy(name = value.asString)
+              case (key @ RespValue.BulkString(_), value @ RespValue.Integer(_))
+                  if XInfoFields.PelCount == key.asString =>
+                readyGroup = readyGroup.copy(pelCount = value.value)
+              case (key @ RespValue.BulkString(_), RespValue.Array(value)) =>
```
anovakovic01, line 443: Please use `values` whenever you're unpacking an iterable.

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -385,6 +385,152 @@ object Output {
...
+                          entryId @ RespValue.BulkString(_),
+                          deliveryTime @ RespValue.Integer(_),
+                          deliveryCount @ RespValue.Integer(_)
+                        )
+                      ) =>
+                    consumerPelList = consumerPelList.:+(
+                      StreamInfoWithFull.ConsumerPel(entryId.asString, deliveryTime.value.millis, deliveryCount.value)
+                    )
+                  case other => throw ProtocolError(s"$other isn't a valid array")
+                }
```
anovakovic01, line 518: I think that you can use `map` here.

redis/src/main/scala/zio/redis/api/Streams.scala
```diff
@@ -48,6 +48,19 @@ trait Streams {
...
+   * Returns the entire state of the stream, including entries, groups, consumers and PELs.
+   *
+   * @param key ID of the stream
+   * @param count limit the amount of stream/PEL entries that are returned (The first <count> entries are returned).
+   * @return General information about the stream stored at the specified key.
+   */
+  final def xInfoStreamFull(
+    key: String,
+    count: Option[Long] = None
+  ): ZIO[RedisExecutor, RedisError, StreamInfoWithFull.FullStreamInfo] =
```
anovakovic01, line 61: You'll have to write `Some` every time you have a count when calling this method. I suggest that you remove a `count` parameter from this method and write another method with the same name that has a `count` parameter of type `Long` in it. That way you wouldn't have to write `Some` every time you need to pass a count.

redis/src/test/scala/zio/redis/InputSpec.scala
```diff
@@ -986,6 +986,13 @@ object InputSpec extends BaseSpec {
           )
         }
       ),
+      suite("XInfoStream FULL")(
```
anovakovic01, line 989: Please rename to `XInfoStreamFull`.

anovakovic01 (changes requested):

redis/src/main/scala/zio/redis/api/Streams.scala
jxnu-liguobin: ok,

jxnu-liguobin: Is it OK to use option internally?(For the input to be generic.)

redis/src/main/scala/zio/redis/Output.scala
jxnu-liguobin: It's going to be more general?

redis/src/main/scala/zio/redis/api/Streams.scala
anovakovic01: Yes, it is.

redis/src/main/scala/zio/redis/Output.scala
anovakovic01: Yes, and the implementation will be cleaner and more consistent because your function will have the extraction logic for only one value.

jxnu-liguobin: done

anovakovic01 (approved):
LGTM!

Issue #325, opened 2021-03-22
Lists API is missing LPOS Command


hcwilhelm:
The Lists API is missing the LPOS Command as specified here https://redis.io/commands/lpos

Issue #324, opened 2021-03-22
Lists API is missing LMOVE command


hcwilhelm:
The List API is missing the LMOVE command as specified here https://redis.io/commands/lmove

hcwilhelm:
I take this one

PR #328, opened 2021-03-23
List api test executor

hcwilhelm:
Closes: https://github.com/zio/zio-redis/issues/195

Support all currently implemented List API Commands including the Blocking commands.

Two commands are missing in the API and thus not included in this PR. See : https://github.com/zio/zio-redis/issues/324 https://github.com/zio/zio-redis/issues/325


redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -1,33 +1,65 @@
...
+
   private def errResponse(cmd: String): RespValue.BulkString =
     RespValue.bulkString(s"(error) ERR wrong number of arguments for '$cmd' command")
 
   private def onConnection(command: String, input: Chunk[RespValue.BulkString])(
     res: => RespValue.BulkString
   ): USTM[BulkString] = STM.succeedNow(if (input.isEmpty) errResponse(command) else res)
 
-  private[this] def runCommand(name: String, input: Chunk[RespValue.BulkString]): STM[RedisError, RespValue] = {
+  private[this] def runCommand(name: String, input: Chunk[RespValue.BulkString]): USTM[RespValue] = {
```
mijicd, line 62: We should make sure no unsafe calls are made, e.g. head on potentially empty chunk. It can (and should) be done in a separate pull request.

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -250,66 +289,361 @@ private[redis] final class TestExecutor private (
...
+            _                <- lists.put(source, sourceResult)
+            _                <- lists.put(destination, destinationResult)
+          } yield RespValue.bulkString(value)).foldM(_ => STM.retry, result => STM.succeed(result))
+        )
+
+      case _ => STM.succeedNow(RespValue.Error("ERR unknown command"))
     }
   }
 
+  private[this] def orWrongType(predicate: ZSTM[Any, Nothing, Boolean])(
```
mijicd, line 614:
```suggestion
  private[this] def orWrongType(predicate: USTM[Boolean])(
```

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -250,66 +289,361 @@ private[redis] final class TestExecutor private (
...
+    program: => ZSTM[Any, Nothing, RespValue]
```
mijicd, line 615:
```suggestion
    program: => USTM[RespValue]
```

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -250,66 +289,361 @@ private[redis] final class TestExecutor private (
...
+  ): ZSTM[Any, Nothing, RespValue] =
```
mijicd, line 616:
```suggestion
  ): USTM[RespValue] =
```

mijicd (commented):
Just a few minor comments, otherwise looks almost ready to go!

redis/src/main/scala/zio/redis/TestExecutor.scala
hcwilhelm: I am not sure if I understand what you are trying to say here. From my point of view, we statically know which inputs we feed into each command and therefore we are able to decode the input even with the use of unsafe operations like `head` etc. How do you imagine that we ensure that `no unsafe calls are made`?

mijicd: What I meant is double-checking if we know statically (last time I took a look there were quite a few head calls), and if now, changing the type to NonEmptyChunk :). If there's nothing to be done here, it's a no-op (even better :) ).

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -1,33 +1,65 @@
...
 
 import zio._
+import zio.duration._
 import zio.redis.RedisError.ProtocolError
 import zio.redis.RespValue.BulkString
-import zio.stm.{ STM, TMap, TRef, USTM }
+import zio.stm.{ random => _, _ }
 
 private[redis] final class TestExecutor private (
+  lists: TMap[String, Vector[String]],
```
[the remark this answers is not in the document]
hcwilhelm: Ok  changed to Chunk

redis/src/main/scala/zio/redis/TestExecutor.scala
hcwilhelm: Alright got it. But as you said this can be done in a follow-up PR.

mijicd (approved):

PR #330, opened 2021-03-24
Implement LMOVE and BLMOVE

hcwilhelm:
Closes: https://github.com/zio/zio-redis/issues/324

Implementation of LMOVE and BLMOVE Commands including tests and Implementation for the TestExecutor

### DIFF
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -97,6 +97,10 @@ object Input {
     def encode(data: Position): Chunk[RespValue.BulkString] = Chunk.single(encodeString(data.stringify))
   }
 
+  case object SideInput extends Input[Side] {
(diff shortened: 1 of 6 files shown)


redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -97,6 +97,10 @@ object Input {
     def encode(data: Position): Chunk[RespValue.BulkString] = Chunk.single(encodeString(data.stringify))
   }
 
+  case object SideInput extends Input[Side] {
```
hcwilhelm, line 100: I need to add a test for this !

mijicd (approved):
Good to go once you add the last remaining test.

redis/src/main/scala/zio/redis/Input.scala
hcwilhelm: Test added

mijicd (approved):

PR #331, opened 2021-03-25
Lpos

hcwilhelm:
Closes: https://github.com/zio/zio-redis/issues/325

Implementation of the List API LPOS command, and tests and support for the TestExecutor


redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -348,6 +348,16 @@ object Input {
     }
   }
 
+  case object ListMaxLenInput extends Input[ListMaxLen] {
+    override private[redis] def encode(data: ListMaxLen): Chunk[RespValue.BulkString] =
```
mijicd, line 352: I'd remove override private prefix, just to align it with othe inputs.

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -184,6 +184,21 @@ object Output {
       }
   }
 
+  case object OptionalLongOrLongArrayOutput extends Output[Either[Option[Long], Chunk[Long]]] {
```
mijicd, line 187: We introduced a convention recently to provide separate commands based on the output so that we don't force pattern match on the call site. Check out geo commands for an example.

redis/src/main/scala/zio/redis/options/Streams.scala
```diff
@@ -75,7 +75,7 @@ trait Streams {
 
   type NoAck = NoAck.type
 
-  case class MaxLen(approximate: Boolean, count: Long)
+  case class StreamMaxLen(approximate: Boolean, count: Long)
```
mijicd, line 78: Not necessary related to this PR, but I'd make these sealed.

redis/src/main/scala/zio/redis/Output.scala
hcwilhelm: Yea I also thought about that but decided to keep it as close to the Redis API as possible. I'll the geo commands and split the LPOS into two commands.

redis/src/main/scala/zio/redis/Input.scala
hcwilhelm: done

redis/src/main/scala/zio/redis/Output.scala
hcwilhelm: Done Split into `Lpos` and `LposCount` commands

redis/src/main/scala/zio/redis/options/Streams.scala
hcwilhelm: Done and made a lot of other case classes sealed. Can you maybe explain to me why the compiler refused to accept `final` case classes here that would even be better than just sealed or do I get something wrong?

mijicd (approved):

Issue #329, opened 2021-03-24
Mission options for ZADD


hcwilhelm:
According to the Redis docs ZADD https://redis.io/commands/zadd ZADD supports `[NX|XX] [GT|LT] [CH] [INCR]` options but the current zio-redis implementation only supports `[NX|XX][CH]` options.

We should implement the missing options and write proper tests for them.

jxnu-liguobin:
Has anyone started?

hcwilhelm:
> Has anyone started?

I didn’t start with that you can take it. 👍

PR #333, opened 2021-03-27
Add `LT` and `GT` options for ZADD command

jxnu-liguobin:
close https://github.com/zio/zio-redis/issues/329

### DIFF
--- a/redis/src/main/scala/zio/redis/options/Shared.scala
+++ b/redis/src/main/scala/zio/redis/options/Shared.scala
@@ -4,14 +4,18 @@ trait Shared {
   sealed trait Update { self =>
     private[redis] final def stringify: String =
       self match {
-        case Update.SetExisting => "XX"
-        case Update.SetNew      => "NX"
+        case Update.SetExisting    => "XX"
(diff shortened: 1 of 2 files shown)


redis/src/main/scala/zio/redis/options/Shared.scala
```diff
@@ -4,14 +4,18 @@ trait Shared {
   sealed trait Update { self =>
     private[redis] final def stringify: String =
       self match {
-        case Update.SetExisting => "XX"
-        case Update.SetNew      => "NX"
+        case Update.SetExisting    => "XX"
+        case Update.SetNew         => "NX"
+        case Update.SetLessThan    => "LT"
+        case Update.SetGreaterThan => "GT"
```
mijicd, line 10: These are independent from XX and NX according to the specification, i.e. we'll need to be able to express both.

mijicd: Ah, scratch that, the docs are misleading.

mijicd (approved):

PR #340, opened 2021-04-06
Support GETEX and GETDEL

### LINKED ISSUE #334: Support `GETEX` and `GETDEL `
see https://redis.io/commands/getex

Available since 6.2.0. 

Using one issue, because two commands are relatively easy(I hope so).

jxnu-liguobin:
closed #334

### DIFF
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -478,12 +478,12 @@ object Input {
       }
   }
 
-  case object GetExAtInput extends Input[(String, Expire, Instant)] {
-    override private[redis] def encode(data: (String, Expire, Instant)): Chunk[RespValue.BulkString] =
+  case object GetExAtInput extends Input[(String, ExpiredAt, Instant)] {
+    override private[redis] def encode(data: (String, ExpiredAt, Instant)): Chunk[RespValue.BulkString] =
       data._2 match {
-        case Expire.SetExpireAtSeconds =>
+        case ExpiredAt.SetExpireAtSeconds =>
           Chunk(encodeString(data._1), encodeString("EXAT")) ++ TimeSecondsInput.encode(data._3)
-        case Expire.SetExpireAtMilliseconds =>
+        case ExpiredAt.SetExpireAtMilliseconds =>
           Chunk(encodeString(data._1), encodeString("PXAT")) ++ TimeMillisecondsInput.encode(data._3)
         case _ => Chunk.empty
       }
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -463,6 +463,32 @@ object Input {
     def encode(data: Update): Chunk[RespValue.BulkString] = Chunk.single(encodeString(data.stringify))
   }
 
+  case object GetExPersistInput extends Input[(String, Boolean)] {
+    override private[redis] def encode(data: (String, Boolean)): Chunk[RespValue.BulkString] =
+      if (data._2) Chunk(encodeString(data._1), encodeString("PERSIST")) else Chunk(encodeString(data._1))
+  }
+  case object GetExInput extends Input[(String, Expire, Duration)] {
+    override private[redis] def encode(data: (String, Expire, Duration)): Chunk[RespValue.BulkString] =
+      data._2 match {
+        case Expire.SetExpireSeconds =>
+          Chunk(encodeString(data._1), encodeString("EX")) ++ DurationSecondsInput.encode(data._3)
+        case Expire.SetExpireMilliseconds =>
+          Chunk(encodeString(data._1), encodeString("PX")) ++ DurationMillisecondsInput.encode(data._3)
+        case _ => Chunk.empty
+      }
+  }
+
+  case object GetExAtInput extends Input[(String, Expire, Instant)] {
+    override private[redis] def encode(data: (String, Expire, Instant)): Chunk[RespValue.BulkString] =
+      data._2 match {
+        case Expire.SetExpireAtSeconds =>
+          Chunk(encodeString(data._1), encodeString("EXAT")) ++ TimeSecondsInput.encode(data._3)
+        case Expire.SetExpireAtMilliseconds =>
+          Chunk(encodeString(data._1), encodeString("PXAT")) ++ TimeMillisecondsInput.encode(data._3)
+        case _ => Chunk.empty
+      }
+  }
+
   final case class Varargs[-A](input: Input[A]) extends Input[Iterable[A]] {
     def encode(data: Iterable[A]): Chunk[RespValue.BulkString] =
       data.foldLeft(Chunk.empty: Chunk[RespValue.BulkString])((acc, a) => acc ++ input.encode(a))
--- a/redis/src/main/scala/zio/redis/Input.scala
+++ b/redis/src/main/scala/zio/redis/Input.scala
@@ -469,22 +469,22 @@ object Input {
   }
   case object GetExInput extends Input[(String, Expire, Duration)] {
     override private[redis] def encode(data: (String, Expire, Duration)): Chunk[RespValue.BulkString] =
-      data._2 match {
-        case Expire.SetExpireSeconds =>
-          Chunk(encodeString(data._1), encodeString("EX")) ++ DurationSecondsInput.encode(data._3)
-        case Expire.SetExpireMilliseconds =>
-          Chunk(encodeString(data._1), encodeString("PX")) ++ DurationMillisecondsInput.encode(data._3)
+      data match {
+        case (key, Expire.SetExpireSeconds, duration) =>
+          Chunk(encodeString(key), encodeString("EX")) ++ DurationSecondsInput.encode(duration)
+        case (key, Expire.SetExpireMilliseconds, duration) =>
+          Chunk(encodeString(key), encodeString("PX")) ++ DurationMillisecondsInput.encode(duration)
         case _ => Chunk(encodeString(data._1))
       }
   }
 
   case object GetExAtInput extends Input[(String, ExpiredAt, Instant)] {
     override private[redis] def encode(data: (String, ExpiredAt, Instant)): Chunk[RespValue.BulkString] =
-      data._2 match {
-        case ExpiredAt.SetExpireAtSeconds =>
-          Chunk(encodeString(data._1), encodeString("EXAT")) ++ TimeSecondsInput.encode(data._3)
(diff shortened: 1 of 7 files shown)


redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -463,6 +463,32 @@ object Input {
...
+
+  case object GetExAtInput extends Input[(String, ExpiredAt, Instant)] {
+    override private[redis] def encode(data: (String, ExpiredAt, Instant)): Chunk[RespValue.BulkString] =
+      data._2 match {
+        case ExpiredAt.SetExpireAtSeconds =>
+          Chunk(encodeString(data._1), encodeString("EXAT")) ++ TimeSecondsInput.encode(data._3)
+        case ExpiredAt.SetExpireAtMilliseconds =>
+          Chunk(encodeString(data._1), encodeString("PXAT")) ++ TimeMillisecondsInput.encode(data._3)
+        case _ => Chunk(encodeString(data._1))
+      }
```
hcwilhelm, line 489: Can we pattern match on the whole `data` tuple? This way we would not need this hard to read `_2`, `_1` etc but it's just cosmetics.

redis/src/main/scala/zio/redis/options/Shared.scala
```diff
@@ -18,6 +19,30 @@ trait Shared {
...
+      self match {
+        case ExpiredAt.SetExpireAtSeconds      => "EXAT"
+        case ExpiredAt.SetExpireAtMilliseconds => "PXAT"
+      }
+  }
+  object ExpiredAt {
+    case object SetExpireAtSeconds      extends ExpiredAt
+    case object SetExpireAtMilliseconds extends ExpiredAt
+  }
+
```
hcwilhelm, line 45: Is this really `shared` or should it be part of the `strings` options?

redis/src/test/scala/zio/redis/StringsSpec.scala
```diff
@@ -1435,6 +1437,92 @@ trait StringsSpec extends BaseSpec {
...
         }
+      ),
+      suite("getEx")(
+        testM("value exists after removing ttl") {
+          for {
+            key    <- uuid
+            value  <- uuid
+            _      <- pSetEx(key, 1000.millis, value)
+            exists <- getEx(key, true)
+            _      <- ZIO.sleep(1010.millis)
```
hcwilhelm, line 1448: This is sleeping for a little longer than a second. I would reduce that by at least 10x to not slow down the tests too much. WDTY?

redis/src/main/scala/zio/redis/options/Shared.scala
jxnu-liguobin: Thanks, I'm not sure, just intuition.

redis/src/test/scala/zio/redis/StringsSpec.scala
jxnu-liguobin: Yes

redis/src/main/scala/zio/redis/Input.scala
jxnu-liguobin: Thanks

redis/src/main/scala/zio/redis/options/Shared.scala
hcwilhelm: I suggest moving this to `string` options if these are only used there, maybe you can double-check. Otherwise, let's keep it like it is.

redis/src/test/scala/zio/redis/StringsSpec.scala
```diff
@@ -1435,6 +1437,92 @@ trait StringsSpec extends BaseSpec {
...
+            res    <- get(key)
+          } yield assert(res.isDefined)(equalTo(true)) && assert(exists)(equalTo(Some(value)))
+        } @@ eventually,
+        testM("not found value when set seconds ttl") {
+          for {
+            key    <- uuid
+            value  <- uuid
+            _      <- set(key, value)
+            exists <- getEx(key, Expire.SetExpireSeconds, 1.second)
+            _      <- ZIO.sleep(1020.millis)
```
hcwilhelm, line 1458: I would reduce the sleep time to something around 10millis

redis/src/test/scala/zio/redis/StringsSpec.scala
```diff
@@ -1435,6 +1437,92 @@ trait StringsSpec extends BaseSpec {
...
+            res    <- get(key)
+          } yield assert(res.isDefined)(equalTo(false)) && assert(exists)(equalTo(Some(value)))
+        } @@ eventually,
+        testM("not found value when set milliseconds ttl") {
+          for {
+            key    <- uuid
+            value  <- uuid
+            _      <- set(key, value)
+            exists <- getEx(key, Expire.SetExpireMilliseconds, 1000.millis)
+            _      <- ZIO.sleep(1020.millis)
```
hcwilhelm, line 1468: here too

redis/src/test/scala/zio/redis/StringsSpec.scala
```diff
@@ -1435,6 +1437,92 @@ trait StringsSpec extends BaseSpec {
...
+            res    <- get(key)
+          } yield assert(res.isDefined)(equalTo(false)) && assert(exists)(equalTo(Some(value)))
+        } @@ eventually,
+        testM("not found value when set seconds timestamp") {
+          for {
+            key    <- uuid
+            value  <- uuid
+            _      <- set(key, value)
+            exists <- getExAt(key, ExpiredAt.SetExpireAtSeconds, Instant.now().plusMillis(1000))
+            _      <- ZIO.sleep(1020.millis)
```
hcwilhelm, line 1478: and here also

redis/src/test/scala/zio/redis/StringsSpec.scala
```diff
@@ -1435,6 +1437,92 @@ trait StringsSpec extends BaseSpec {
...
+            res    <- get(key)
+          } yield assert(res.isDefined)(equalTo(false)) && assert(exists)(equalTo(Some(value)))
+        } @@ eventually,
+        testM("not found value when set milliseconds timestamp") {
+          for {
+            key    <- uuid
+            value  <- uuid
+            _      <- set(key, value)
+            exists <- getExAt(key, ExpiredAt.SetExpireAtMilliseconds, Instant.now().plusMillis(1000))
+            _      <- ZIO.sleep(1020.millis)
```
hcwilhelm, line 1488: and here also

redis/src/test/scala/zio/redis/StringsSpec.scala
```diff
@@ -1435,6 +1437,92 @@ trait StringsSpec extends BaseSpec {
...
+            res    <- get(key)
+          } yield assert(res.isDefined)(equalTo(false)) && assert(exists)(equalTo(Some(value)))
+        } @@ eventually,
+        testM("key not found") {
+          for {
+            key   <- uuid
+            value <- uuid
+            _     <- set(key, value)
+            res   <- getExAt(value, ExpiredAt.SetExpireAtMilliseconds, Instant.now().plusMillis(1000))
+            res2  <- getEx(value, Expire.SetExpireMilliseconds, 1000.millis)
```
hcwilhelm, line 1498: and here also

redis/src/test/scala/zio/redis/StringsSpec.scala
```diff
@@ -1435,6 +1437,92 @@ trait StringsSpec extends BaseSpec {
...
+        testM("value exists after removing ttl") {
+          for {
+            key    <- uuid
+            value  <- uuid
+            _      <- pSetEx(key, 10.millis, value)
+            exists <- getEx(key, true)
+            _      <- ZIO.sleep(20.millis)
+            res    <- get(key)
+          } yield assert(res.isDefined)(equalTo(true)) && assert(exists)(equalTo(Some(value)))
+        } @@ eventually,
```
hcwilhelm, line 1451: Just a question why is this test `@eventually`?

redis/src/test/scala/zio/redis/StringsSpec.scala
jxnu-liguobin: The problem here is that the minimum expiration time is one second.

redis/src/test/scala/zio/redis/StringsSpec.scala
jxnu-liguobin: Yes, This can be changed.

redis/src/test/scala/zio/redis/StringsSpec.scala
hcwilhelm: Ok I see, thanks for clarifying

redis/src/test/scala/zio/redis/StringsSpec.scala
jxnu-liguobin: This is because sleep is unreliable and subject to CPU and thread scheduling. 20ms is likely to cause failure. Here, it is similar to CAS. (guess)

hcwilhelm: Ah true, I forgot that we use the Live Clock in these tests.

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -465,29 +465,33 @@ object Input {
       Chunk.single(encodeString(data.stringify))
   }
 
-  case object GetExPersistInput extends Input[(String, Boolean)] {
-    override private[redis] def encode(data: (String, Boolean)): Chunk[RespValue.BulkString] =
-      if (data._2) Chunk(encodeString(data._1), encodeString("PERSIST")) else Chunk(encodeString(data._1))
-  }
-  case object GetExInput extends Input[(String, Expire, Duration)] {
-    override private[redis] def encode(data: (String, Expire, Duration)): Chunk[RespValue.BulkString] =
+  case class GetExPersistInput[K: Schema]() extends Input[(K, Boolean)] {
```
mijicd, line 468: Case classes should be final.

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -465,29 +465,33 @@ object Input {
...
+    override private[redis] def encode(data: (K, Boolean))(implicit codec: Codec): Chunk[RespValue.BulkString] =
```
mijicd, line 469: Let's remove `override private`. I noticed that some of the recently added inputs are using it, but please remove it everywhere together with the ones from this PR.

hcwilhelm:
Please mark all test wich use generators eg `checkM(...) { }` with `@@ flaky`

hcwilhelm:
👍

mijicd (approved):

PR #354, opened 2021-04-16
add zMScore

### LINKED ISSUE #351: Missing Command ZMSCORE
The Redis command ZMSCORE is missing in the sortedSet API https://redis.io/commands/zmscore

jxnu-liguobin:
close #351

Redis doc is `Array reply: list of scores or nil associated with the specified member values (a double precision floating point number), represented as strings.`
`Chunk[Option[Double]]]` is used, but I don't think it's the most suitable one. Does anyone have a better proposal?

### DIFF
--- a/redis/src/test/scala/zio/redis/SortedSetsSpec.scala
+++ b/redis/src/test/scala/zio/redis/SortedSetsSpec.scala
@@ -986,6 +986,27 @@ trait SortedSetsSpec extends BaseSpec {
           } yield assert(result)(isNone)
         }
       ),
+      suite("zMScore")(
+        testM("non-empty set") {
+          for {
+            key <- uuid
+            _ <- zAdd(key)(
+                   MemberScore(10d, "Delhi"),
+                   MemberScore(20d, "Mumbai"),
+                   MemberScore(30d, "Hyderabad"),
+                   MemberScore(40d, "Kolkata"),
+                   MemberScore(50d, "Chennai")
+                 )
+            result <- zMScore(key, "Delhi", "Mumbai")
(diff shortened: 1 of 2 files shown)


hcwilhelm:
> close #351
> 
> Redis doc is `Array reply: list of scores or nil associated with the specified member values (a double precision floating point number), represented as strings.`
> `Chunk[Option[Double]]]` is used, but I don't think it's the most suitable one. Does anyone have a better proposal?

I think this is totally fine as it reflects the Redis response. `For every member that does not exist in the sorted set, a nil value is returned.`

👍

redis/src/test/scala/zio/redis/SortedSetsSpec.scala
```diff
@@ -986,6 +986,27 @@ trait SortedSetsSpec extends BaseSpec {
...
+            key <- uuid
+            _ <- zAdd(key)(
+                   MemberScore(10d, "Delhi"),
+                   MemberScore(20d, "Mumbai"),
+                   MemberScore(30d, "Hyderabad"),
+                   MemberScore(40d, "Kolkata"),
+                   MemberScore(50d, "Chennai")
+                 )
+            result <- zMScore(key, "Delhi", "Mumbai")
+          } yield assert(result)(equalTo(Chunk(Some(10d), Some(20d))))
```
mijicd, line 1001: I'd include the "non-existing" value here.

jxnu-liguobin: good case 👍

mijicd (approved):

PR #355, opened 2021-04-19
fix issues that methods does not work with protobuf codec

anatolysergeev:


redis/src/test/scala/zio/redis/BaseSpec.scala
```diff
@@ -17,7 +16,7 @@ trait BaseSpec extends DefaultRunnableSpec {
 
   def instantOf(millis: Long): UIO[Instant] = UIO(Instant.now().plusMillis(millis))
 
-  implicit val codec: Codec = StringUtf8Codec
+  implicit val codec: Codec = ProtobufCodec
```
jxnu-liguobin, line 19: Could you tell me why we need to use `Protobufcodec`? Thank you.

hcwilhelm: Yea I am also curious

mijicd: I presume you're referring to the test usage, not the general motivation :). That said, we don't need to use it in the tests, but @anatolysergeev realized there were some discrepancies with it. To be more precise, the expectation is that the library should work properly given any compatible codec, which wasn't the case here.

anatolysergeev: Yeah, for example if we would encode int, long, double with protobuf codec, we would get some extra bytes that would describe a type of our value, but redis don't understand this types, and how to work with that, redis just want to see double (for example). That's why the idea to use ArbitraryType[Long] in some cases was a bad idea. If we use something else except StringUtf8 codec, that could broke contract with redis.
If we use in tests for example protobuf codec, we are gonna see that bound whether we can use ArbitraryType or not. May be we should consider the both codecs in tests, because, i'm afraid that if we would use only protobuf codec than we can miss something with our simple codec.)
Here i just want to show that now all tests are working with protobuf codec. I can change it to String codec now, should i?)

mijicd: Based on the information from discord regarding the readiness of protobuf codec, I'd say let's postpone on using it in the tests for now. Later on we may fully switch to it, or run them together.

anatolysergeev: Ok. I've returned string codec to tests.

mijicd (approved):

PR #363, opened 2021-04-29
Test executor sortedsets

hcwilhelm:
SortedSets API support for the TestExecutor backed by a Map[Element, Score]
I found it easier to have a Map of Element -> Score as the underlying data structure than for example a scala TreeSet.

The implementation is not complete at this point but I wanted to create a PR anyways to get some feedback. 
I am struggling with implementing full SortedSet API support because some commands support an enormous amount of 
Options and also for some commands It is not directly obvious how they should work.

### DIFF
--- a/redis/src/main/scala/zio/redis/TestExecutor.scala
+++ b/redis/src/main/scala/zio/redis/TestExecutor.scala
@@ -988,6 +988,49 @@ private[redis] final class TestExecutor private (
           } yield RespValue.Array(Chunk.fromIterable(values))
         )
 
+      case api.Hashes.HRandField =>
+        val key        = input(0).asString
+        val count      = if (input.size == 1) None else Some(input(1).asString.toLong)
+        val withValues = if (input.size == 3) input(2).asString == "WITHVALUES" else false
+
+        def selectValues[T](n: Long, values: Vector[T]) = {
+          val repeatedAllowed = n < 0
+          val count           = Math.abs(n)
+          val t               = count - values.length
+
+          if (repeatedAllowed && t > 0) {
+            selectNWithReplacement[T](values, count, randomPick)
+          } else {
+            selectN[T](values, count, randomPick)
+          }
+        }
+
+        orWrongType(isHash(key)) {
+          val keysAndValues = for {
+            hash <- hashes.getOrElse(key, Map.empty)
+          } yield (hash.keys map RespValue.bulkString) zip (hash.values map RespValue.bulkString)
+
+          if (count.isDefined && withValues) {
+            for {
+              kvs            <- keysAndValues
+              fields         <- selectValues(count.get, kvs.toVector)
+              fieldsAndValues = fields.flatMap { case (k, v) => Seq(k, v) }
+            } yield RespValue.Array(Chunk.fromIterable(fieldsAndValues))
+          } else if (count.isDefined) {
+            for {
+              kvs    <- keysAndValues
+              keys    = kvs.map(_._1)
+              fields <- selectValues(count.get, keys.toVector)
+            } yield RespValue.Array(Chunk.fromIterable(fields))
+          } else {
+            for {
+              kvs <- keysAndValues
+              keys = kvs.map { case (k, _) => k }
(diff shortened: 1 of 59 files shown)


redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -972,6 +980,972 @@ private[redis] final class TestExecutor private (
...
+//            newSet         = ss.drop(1)
+//            _             <- sortedSets.put(sk, newSet)
+//          } yield Replies.array(Chunk(sk, min.member, min.score.toString)))
+//            .foldM(_ => STM.retry, result => STM.succeed(result))
+//        )
+//
+      case api.SortedSets.ZAdd =>
+        val key = input(0).asString
+
+        val updateOption = input.map(_.asString).find {
```
mijicd, line 1018: Does it need conversion to string? Also, `|` might make it cleaner:

```scala
case "XX" | "NX" | "LT" | "GT" => true
case _ => false
```

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -972,6 +980,972 @@ private[redis] final class TestExecutor private (
...
+        }
+
+        val incrOption = input.map(_.asString).find {
+          case "INCR" => true
+          case _      => false
+        }
+
+        val optionsCount = updateOption.map(_ => 1).getOrElse(0) + changedOption.map(_ => 1).getOrElse(0) + incrOption
+          .map(_ => 1)
+          .getOrElse(0)
```
mijicd, line 1038: Please use `fold` instead of `map` + `getOrElse`. This remark applies generally :).

mijicd (commented):
I'll make another, more detailed pass as I need to review this using editor to understand everything, but in general, mapping doesn't seem to be troublesome.

redis/src/test/scala/zio/redis/ApiSpec.scala
```diff
@@ -32,13 +32,14 @@ object ApiSpec
...
-        connectionSuite,
-        setsSuite,
-        hyperLogLogSuite,
-        listSuite,
-        hashSuite
+        connectionSuite @@ TestAspect.ignore,
+        setsSuite @@ TestAspect.ignore,
+        hyperLogLogSuite @@ TestAspect.ignore,
+        listSuite @@ TestAspect.ignore,
+        hashSuite @@ TestAspect.ignore,
```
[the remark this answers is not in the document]
hcwilhelm: I ignored them only to run the tests for the sorted set API. That helped me to unclutter the whole test output. I will enable them of course when this PR is in the final state.

hcwilhelm:
> I'll make another, more detailed pass as I need to review this using editor to understand everything, but in general, mapping doesn't seem to be troublesome.

Ok, that's good. I find it way easier to work with the map where I have access to both the value and the score. And so far I could satisfy all the tests I allowed to run ;) And yes I fully understand this is a giant PR and the sorted set Implementation has by far the most lines of code, although I am sure there is a lost of duplication.

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -2087,11 +2088,60 @@ private[redis] final class TestExecutor private (
...
+        val key              = input(0).asString
+        val maybeCount        = input.tail.headOption.map(b => b.asString.toLong)
+        val withScoresOption = input.map(_.asString).find(_ == "WITHSCORES")
+
+        orWrongType(isSortedSet(key))(
+          for {
+            scoreMap <- sortedSets.getOrElse(key, Map.empty)
+            asVector = scoreMap.toVector
+            res <- maybeCount match {
+              case None => selectOne(asVector, randomPick).map {
```
jxnu-liguobin, line 2106: Is `selectOne` method provided by ZIO?

hcwilhelm: @jxnu-liguobin no it's defined in the same file on line 2229

mijicd (approved):

PR #365, opened 2021-05-03
Set explicit container version

rfmejia:
In order to test new commands introduced in Redis 6.2, upgrade our test container versions.


mijicd (approved):
These two tags are referring to the same redis version (6.2.3), so technically this PR isn't doing much. However, I'm merging it to avoid further confusion.

Issue #342, opened 2021-04-08
Missing Command HRANDFIELD


hcwilhelm:
According to https://redis.io/commands/hrandfield Redis support the command HRANDFIELD, but this command is missing in the Hashes API.

## Tasks ##

- [ ] Implement the command
- [ ] Add tests for the command
- [ ] Add support for this command to the TestExecutor

jgoday:
@hcwilhelm I have try to implement this with this PR #370.
I have implemented two different methods depending on the return value (field or array of fields).
Is this a valid approach ?

hcwilhelm:
@jgoday i will review it in the next couple of days. But from what you described this approach sounds right to me. We split out different methods depending on the return type.

PR #370, opened 2021-05-07
#342: Implement hrandfield command and support it in TestExecutor

jgoday:
Implements HRANDFIELD as two different methods.
As mentioned in the redis documentation, HRANGEFIELD can return a field or an array of fields/values, depending on the value of the count parameter.
This implementation has two different methods, differentiated by their return value.
If count is negative, it allows to return repeated field/values.

* hRandField[K: Schema, V: Schema](key: K): ZIO[RedisExecutor, RedisError, Option[V]]
* hRandField[K: Schema, V: Schema](
    key: K,
    count: Long,
    withValues: Boolean = false
  ): ZIO[RedisExecutor, RedisError, Chunk[V]]

### DIFF
--- a/redis/src/main/scala/zio/redis/TestExecutor.scala
+++ b/redis/src/main/scala/zio/redis/TestExecutor.scala
@@ -972,6 +972,49 @@ private[redis] final class TestExecutor private (
           } yield RespValue.Array(Chunk.fromIterable(values))
         )
 
+      case api.Hashes.HRandField =>
+        val key        = input(0).asString
+        val count      = if (input.size == 1) None else input(1).asString.toLongOption
+        val withValues = if (input.size == 3) input(2).asString == "WITHVALUES" else false
+
+        def selectValues[T](n: Long, values: Vector[T]) = {
+          val repeatedAllowed = n < 0
+          val count           = Math.abs(n)
+          val t               = count - values.length
+
+          if (repeatedAllowed && t > 0) {
+            selectNWithReplacement[T](values, count, randomPick)
+          } else {
+            selectN[T](values, count, randomPick)
+          }
+        }
+
+        orWrongType(isHash(key)) {
+          val keysAndValues = for {
+            hash <- hashes.getOrElse(key, Map.empty)
+          } yield (hash.keys map RespValue.bulkString) zip (hash.values map RespValue.bulkString)
+
+          if (count.isDefined && withValues) {
+            for {
+              kvs            <- keysAndValues
+              fields         <- selectValues(count.get, kvs.toVector)
+              fieldsAndValues = fields.map { case (k, v) => Seq(k, v) }.flatten
+            } yield RespValue.Array(Chunk.fromIterable(fieldsAndValues))
+          } else if (count.isDefined) {
+            for {
+              kvs    <- keysAndValues
+              keys    = kvs.map(_._1)
+              fields <- selectValues(count.get, keys.toVector)
+            } yield RespValue.Array(Chunk.fromIterable(fields))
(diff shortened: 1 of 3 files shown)


redis/src/main/scala/zio/redis/api/Hashes.scala
```diff
@@ -240,6 +240,41 @@ trait Hashes {
...
+   * @param key of the hash which fields should be read
+   * @return random field in the hash or None when `key` does not exist.
+   */
+  final def hRandField[K: Schema, V: Schema](key: K): ZIO[RedisExecutor, RedisError, Option[V]] = {
+    val command = RedisCommand(HRandField, ArbitraryInput[K](), OptionalOutput(ArbitraryOutput[V]()))
+    command.run(key)
+  }
+
+  /**
+   * Return an array of distinct fields. The array's length is either `count` or the hash's number of fields if `count` is greater.
```
jxnu-liguobin, line 256: Please keep the document consistent with other commands.

jxnu-liguobin: 1. @param key of a sorted set
2. @param key Key of a sorted set 

There are still two styles of documentation, but I'm not sure which is the standard(Maybe the second one? Scala source code uses the second one.). An additional PR may be needed to unify all document comments.

mijicd: Definitely the second one. However, if there are no proboems with the implementation, let's merge this and do the docs unification in the follow up.

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -972,6 +972,49 @@ private[redis] final class TestExecutor private (
...
+        orWrongType(isHash(key)) {
+          val keysAndValues = for {
+            hash <- hashes.getOrElse(key, Map.empty)
+          } yield (hash.keys map RespValue.bulkString) zip (hash.values map RespValue.bulkString)
+
+          if (count.isDefined && withValues) {
+            for {
+              kvs            <- keysAndValues
+              fields         <- selectValues(count.get, kvs.toVector)
+              fieldsAndValues = fields.map { case (k, v) => Seq(k, v) }.flatten
```
hcwilhelm, line 1001: Instead of `flatten` it should be possible to directly use `flatMap`

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -972,6 +972,49 @@ private[redis] final class TestExecutor private (
...
+            } yield RespValue.Array(Chunk.fromIterable(fieldsAndValues))
+          } else if (count.isDefined) {
+            for {
+              kvs    <- keysAndValues
+              keys    = kvs.map(_._1)
```
hcwilhelm, line 1006: For better readability, I prefer `kvs.map {case (k, _) => k}` but up to you just a suggestion.

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -972,6 +972,49 @@ private[redis] final class TestExecutor private (
           } yield RespValue.Array(Chunk.fromIterable(values))
         )
 
+      case api.Hashes.HRandField =>
+        val key        = input(0).asString
+        val count      = if (input.size == 1) None else input(1).asString.toLongOption
```
hcwilhelm, line 977: I think `.toLongOption` is not available in scala 2.12

hcwilhelm:
LGTM

mijicd (approved):

PR #404, opened 2021-06-14
Improve polymorphic output handling

mijicd:
This pull request improves library usability by removing the need to specify explicit input and output types.

### DIFF
--- a/redis/src/main/scala/zio/redis/api/Keys.scala
+++ b/redis/src/main/scala/zio/redis/api/Keys.scala
@@ -103,10 +103,11 @@ trait Keys {
    * @return
    *   keys matching pattern.
    */
-  final def keys[V: Schema](pattern: String): ZIO[RedisExecutor, RedisError, Chunk[V]] = {
-    val command = RedisCommand(Keys.Keys, StringInput, ChunkOutput(ArbitraryOutput[V]()))
-    command.run(pattern)
-  }
+  final def keys(pattern: String): ResultBuilder[Chunk] =
+    new ResultBuilder[Chunk] {
+      def returning[V: Schema]: ZIO[RedisExecutor, RedisError, Chunk[V]] =
+        RedisCommand(Keys.Keys, StringInput, ChunkOutput(ArbitraryOutput[V]())).run(pattern)
+    }
 
   /**
    * Atomically transfer a key from a source Redis instance to a destination Redis instance. On success the key is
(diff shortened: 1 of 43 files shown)


redis/src/main/scala/zio/redis/api/Keys.scala
```diff
@@ -344,19 +347,22 @@ trait Keys {
...
     `type`: Option[RedisType] = None
-  ): ZIO[RedisExecutor, RedisError, (Long, Chunk[K])] = {
-    val command = RedisCommand(
-      Scan,
-      Tuple4(LongInput, OptionalInput(PatternInput), OptionalInput(CountInput), OptionalInput(RedisTypeInput)),
-      Tuple2Output(ArbitraryOutput[Long](), ChunkOutput(ArbitraryOutput[K]()))
-    )
-    command.run((cursor, pattern.map(Pattern), count, `type`))
-  }
+  ): ResultBuilder1[({ type lambda[x] = (Long, Chunk[x]) })#lambda] =
```
anovakovic01, line 355: Would it make sense to declare a type alias (for example called `Cursor`) instead of using type lambda?

redis/src/main/scala/zio/redis/api/SortedSets.scala
```diff
@@ -842,22 +897,24 @@ trait SortedSets {
...
-      ChunkTuple2Output(ArbitraryOutput[M](), DoubleOutput)
-        .map(_.map { case (m, s) => MemberScore(s, m) })
-    val command = RedisCommand(
-      ZScan,
-      Tuple4(ArbitraryInput[K](), LongInput, OptionalInput(PatternInput), OptionalInput(CountInput)),
-      Tuple2Output(MultiStringOutput.map(_.toLong), memberScoresOutput)
-    )
-    command.run((key, cursor, pattern.map(Pattern), count))
-  }
+  ): ResultBuilder1[({ type lambda[x] = (Long, MemberScores[x]) })#lambda] =
```
anovakovic01, line 905: Can you create a type alias instead of using type lambda?

anovakovic01 (changes requested):
This will be a great usability improvement, I really like it! Good job!

redis/src/main/scala/zio/redis/api/Keys.scala
mijicd: We have to use type lambdas here. An alternative is to use the kind-projector, which we discussed [here](https://github.com/mijicd/zio-redis/pull/1#discussion_r777146269).

redis/src/main/scala/zio/redis/api/SortedSets.scala
mijicd: Same here.

redis/src/main/scala/zio/redis/api/Keys.scala
mijicd: If you're referring to the possibility of defining a "top-level" alias, and using it, while it is possible, I'd say it's not bringing much to the plate as it would be used only here, so I'd rather stick to the current approach.

anovakovic01: Yes, I was referring to the possibility of defining a top-level alias. Sure, leave it like it is then.

anovakovic01 (dismissed):
LGTM!

anovakovic01 (approved):

adamgfraser (approved):

PR #371, opened 2021-05-09
Implement keys API in test executor (#194)

rfmejia:
This draft PR implements test executors for keys (#194) in two parts. Please comment on the approach if you have any concerns/improvements:

### Key Management
All keys are tracked in `keys: TSet[String]`. Support for key-related commands are performed against this set. To modify an underlying data structure, Redis commands must use functions that also manage adding keys and expirations (e.g., `def putList = list.put(...) <* keys.put(...) <* expirations.delete(...)`).

### Expiration Management
We introduce the concept of time whenever a Redis command is executed. First, keys are mapped to its expiration Instant (i.e., *when* it expires) in `expirations: TMap[String, Instant]`. Mappings are added using EXPIRE, EXPIREAT, etc. Mappings are removed using PERSIST or any command that modifies an existing key.

When a command is executed, the current time is captured from `zio.clock.Clock`. All keys in `expirations` with an instant before the current time are purged. Finally, execute the Redis command as usual using put and delete functions.

#### Action items
Implement the rest of commands in the keys test suite. Also implement commands in other test suites that use expirations (e.g., PSETEX). Finally, figure out how to use `zio.test.environment.TestClock` for reliable testing of expirations.

Issue #194, opened 2020-11-21
Add keys api to test executor


mschuwalow:
The [test executor](https://github.com/zio/zio-redis/blob/master/redis/src/main/scala/zio/redis/Interpreter.scala#L80) currently only supports a subset of the api.
The goal of this issue is to support the full keys api in the test executor. Tests for this group should be enabled for the test executor [here](https://github.com/zio/zio-redis/blob/master/redis/src/test/scala/zio/redis/ApiSpec.scala#L32)

mijicd:
Closed in #480

PR #480, opened 2021-11-09
Implement keys API in test executor #194

anatolysergeev:
This PR is continuation of PR https://github.com/zio/zio-redis/pull/371
Some changes in the concept how we save expiration time, before there was an idea to have two separated stores for keys and expiration time, I do not very like that idea because it doesn't have good scale options. I decided to unite them in one key store where we save all information about our key: type, expiration time, etc.

### DIFF
--- a/redis/src/main/scala/zio/redis/TestExecutor.scala
+++ b/redis/src/main/scala/zio/redis/TestExecutor.scala
@@ -1,6 +1,6 @@
 package zio.redis
 
-import java.nio.file.{FileSystems, Path}
+import java.nio.file.{FileSystems, Paths}
 import java.time.Instant
 
 import zio._
@@ -676,7 +676,7 @@ private[redis] final class TestExecutor private (
         val pattern = input.head.asString
         val matcher = FileSystems.getDefault.getPathMatcher("glob:" + pattern)
         keys.keys
-          .map(keys => keys.filter(k => matcher.matches(Path.of(k))))
+          .map(keys => keys.filter(k => matcher.matches(Paths.get(k))))
           .map(Replies.array)
 
       case api.Keys.Scan =>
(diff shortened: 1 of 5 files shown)


redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -650,6 +660,193 @@ private[redis] final class TestExecutor private (
...
+        keys.keys
+          .map(keys => keys.filter(k => matcher.matches(Path.of(k))))
+          .map(Replies.array)
+
+      case api.Keys.Scan =>
+        val start   = input.head.asString.toInt
+        val options = input.drop(1).map(_.asString)
+
+        val countIndex = options.indexOf("COUNT")
+        val count      = Option.when(countIndex >= 0)(options(countIndex + 1).toInt)
```
mijicd, line 689: I think that `Option.when` won't work in 2.12.

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -650,6 +660,193 @@ private[redis] final class TestExecutor private (
...
+      case api.Keys.Persist =>
+        val key = input.head.asString
+        keys
+          .get(key)
+          .flatMap(_.fold(STM.succeedNow(0L))(info => keys.put(key, info.copy(expireAt = None)).as(1L)))
+          .map(RespValue.Integer)
+
+      case api.Keys.Expire =>
+        val key      = input(0).asString
+        val unixtime = now.plusSeconds(input(1).asLong)
```
mijicd, line 820: This should be suspended, but since we don't have a proper solution for clock in stm, it can stay.

redis/src/main/scala/zio/redis/TestExecutor.scala
```diff
@@ -1351,7 +1564,8 @@ private[redis] final class TestExecutor private (
             hash       <- hashes.getOrElse(key, Map.empty)
             countExists = hash.keys count values.contains
             newHash     = hash -- values
-            _          <- if (newHash.isEmpty) hashes.delete(key) else hashes.put(key, newHash)
+            _ <- if (newHash.isEmpty) delete(key)
+                 else putHash(key, newHash)
```
mijicd, line 1568: You should be able to use `ZIO.cond` for all similar occurrences.

mijicd (commented):
A few remarks, other than that it looks great!

mijicd (approved):

PR #557, opened 2022-02-23
Use existing codecs in tests and examples

mijicd:
This pull request update zio-schema dependency to the latest stable version.

Closes #553
Closes https://github.com/zio/zio-schema/issues/30


redis/src/test/scala/zio/redis/KeysSpec.scala
```diff
@@ -389,25 +388,25 @@ trait KeysSpec extends BaseSpec {
...
           } yield assert(sorted)(equalTo(Chunk("A")))
         },
-        testM("getting multiple value referenced by a key-value pair") {
+        testM("getting multiple values referenced by a key-value pair") {
           for {
             key     <- uuid
-            value1   = 1
-            value2   = 2
+            value1  <- uuid
+            value2  <- uuid
```
mijicd, line 398: Once I set the protobuf codec as the default one in tests, this test started failing. When I tweaked it to use the list of strings instead of the list of integers, it started passing, but it became flaky. Unless @thinkharderdev is aware of some bug that could cause this odd behavior, I would assume that we need to dig into our Redis responses stream, in which case I would address it in the follow-up PR.

anovakovic01 (approved):

redis/src/test/scala/zio/redis/KeysSpec.scala
thinkharderdev: Nothing that I know of.

PR #630, opened 2022-08-21
Fix bulkstring decoder when it ends with a '\n' character

### LINKED ISSUE #613: BulkString with newline character at the end causes process to lock
If a bulkString has a '\n' at the end of the actual string data, RespValue Decoder produces a lock in the current RedisExecutor.

One way to reproduce it is using PING command that returns a BulkString if a string argument is provided.
```scala
testM("PING with a string argument will lock executor") {
  ping(Some("Hello with a newline\n")).map(assert(_)(equalTo("Hello with a newline\n")))
}
```
Executing it (bloop test redis-test -- -t "PING with a string argument will lock executor"),
will cause a timeout exception (Timeout of 1 m exceeded.)

I was facing this issue trying to implement #612 (decoding a client info response returned from the server as a bulk string, client info data has a newline char at the end of the record).

I think the problem is located at RespValue Decoder
```scala
Transducer.utf8Decode >>> Transducer.splitLines >>> processLine
```

Transduder.splitLines will produce an empty line after the bulk string data,
causing an incosistent state in the feed method,
when the bulk string data is completed, state becomes 'Done' and then, with the new empty line, it becomes 'Failed'.

A simple solution seems to have a transition from 'Start' to 'Start' when a empty line is feeded.

Could make a PR if it makes sense.

jgoday:
Fixes #613.

As described in #613, If a bulkString has a '\n' at the end of the actual string data, RespValue Decoder produces a lock in the current RedisExecutor.

This simple PR adds a transition from Start to Start, when the line is blank, in RespValue.feed.

### DIFF
--- a/redis/src/main/scala/zio/redis/RespValue.scala
+++ b/redis/src/main/scala/zio/redis/RespValue.scala
@@ -84,7 +84,7 @@ object RespValue {
         case other             => ZIO.dieMessage(s"Deserialization bug, should not get $other")
       }
 
-    (ZPipeline.utf8Decode >>> ZPipeline.splitLines)
+    (ZPipeline.utf8Decode >>> ZPipeline.splitOn(internal.CrLfString))
       .mapError(e => RedisError.ProtocolError(e.getLocalizedMessage))
       .andThen(ZPipeline.fromSink(lineProcessor))
   }
@@ -105,6 +105,7 @@ object RespValue {
     }
 
     final val CrLf: Chunk[Byte]              = Chunk('\r', '\n')
+    final val CrLfString: String             = "\r\n"
     final val NullArrayEncoded: Chunk[Byte]  = Chunk.fromArray("*-1\r\n".getBytes(StandardCharsets.US_ASCII))
     final val NullArrayPrefix: String        = "*-1"
     final val NullStringEncoded: Chunk[Byte] = Chunk.fromArray("$-1\r\n".getBytes(StandardCharsets.US_ASCII))
@@ -121,6 +122,7 @@ object RespValue {
 
       final def feed(line: String): State =
         self match {
+          case Start if line.isEmpty()           => Start
           case Start if line == NullStringPrefix => Done(NullBulkString)
           case Start if line == NullArrayPrefix  => Done(NullArray)
 
@@ -129,7 +131,9 @@ object RespValue {
               case Headers.SimpleString => Done(SimpleString(line.tail))
               case Headers.Error        => Done(Error(line.tail))
               case Headers.Integer      => Done(Integer(unsafeReadLong(line, 1)))
-              case Headers.BulkString   => ExpectingBulk
+              case Headers.BulkString =>
+                val size = unsafeReadLong(line, 1).toInt
+                CollectingBulkString(size, new StringBuilder(size))
               case Headers.Array =>
                 val size = unsafeReadLong(line, 1).toInt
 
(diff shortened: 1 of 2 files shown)


jgoday:
It seems that the error/locking remains when a bulk string has a '\n' character in a position other than the end.
I will try to fixit tomorrow

mijicd:
@jgoday Thank you. I have marked the PR as draft for now.

jgoday:
Updated to replace RespValue.State ExpectingBulkString to CollectingBulkString,
allowing to gather whole bulk string across several lines.

mijicd:
@jgoday Ready for review?

jgoday:
> @jgoday Ready for review? If so, please convert the draft to PR 🥂

Sure, just converted it to PR again.

redis/src/main/scala/zio/redis/RespValue.scala
```diff
@@ -146,16 +150,24 @@ object RespValue {
...
 
-          case ExpectingBulk => Done(bulkString(line))
-          case _             => Failed
+          case CollectingBulkString(rem, vals, index) =>
+            val allLineBytes = Chunk.fromArray(line.getBytes(StandardCharsets.UTF_8))
+            val lineBytes    = if (allLineBytes.length > rem) allLineBytes.slice(0, rem) else allLineBytes
+            val bytesBuilder = if (index > 0) vals ++= (Lf +: lineBytes) else vals ++= lineBytes
+
+            if (rem - lineBytes.length > 0)
+              CollectingBulkString(rem - lineBytes.length - 1, bytesBuilder, index + lineBytes.length)
```
mijicd, line 159: I'd factor out these length calls.

redis/src/main/scala/zio/redis/RespValue.scala
```diff
@@ -146,16 +150,24 @@ object RespValue {
               case state              => CollectingArray(rem, vals, state.feed)
             }
 
-          case ExpectingBulk => Done(bulkString(line))
-          case _             => Failed
+          case CollectingBulkString(rem, vals, index) =>
+            val allLineBytes = Chunk.fromArray(line.getBytes(StandardCharsets.UTF_8))
```
mijicd, line 154: I'm wondering if it's possible to postpone (or limit) the UTF8 encode & decode, as it will be quite expensive. :thinking:

mijicd (commented):
I suspect that this issue might be present "universally", not only for bulk strings (i.e. presence of newlines in 'content'). Perhaps we can expand the test suite or discard that option.

redis/src/test/scala/zio/redis/ConnectionSpec.scala
```diff
@@ -159,6 +159,13 @@ trait ConnectionSpec extends BaseSpec {
         },
         testM("PING with input") {
           ping(Some("Hello")).map(assert(_)(equalTo("Hello")))
+        },
+        testM("PING with a string argument will not lock executor") {
+          ping(Some("Hello with a newline\n")).map(assert(_)(equalTo("Hello with a newline\n")))
```
anatolysergeev, line 164: Is everything going to be ok if in the end would be \r ?

jxnu-liguobin:
> I suspect that this issue might be present "universally", not only for bulk strings (i.e. presence of newlines in 'content'). Perhaps we can expand the test suite or discard that option.

The issue I created earlier may also be because of this?

redis/src/main/scala/zio/redis/RespValue.scala
jgoday: I have updated it to simplify the CollectinngBulkString logic.
  As stated in redis RESP spec about arrays, I think we can replace Transduce.splitLines for Transduce.splitOn(CrLf),
 so unnecessary complications are avoided when it comes to recomposing the possible value of the bulk string.

redis/src/main/scala/zio/redis/RespValue.scala
jgoday: We can append bulk string lines using a stringbuilder and then, only at the end encoding as UTF bytes.

redis/src/test/scala/zio/redis/ConnectionSpec.scala
jgoday: You are right, as mentioned early, if we split only crlf, then we can recompose the exact value of the bulk string, avoiding an ambiguity between Lf and CrLf

mijicd:
@jgoday Please resolve conflicts so that we can proceed with review & (hopefully :) ) merge :pray:

jgoday:
> @jgoday Please resolve conflicts so that we can proceed with review & (hopefully :) ) merge pray

Sorry again for the delay,
I just have merge and squash it.

mijicd (approved):

PR #653, opened 2022-10-30
Cluster support

### LINKED ISSUE #151: Implement cluster API

### Commands

- [x] CLUSTER ASK When a cluster client receives an -ASK redirect, command is sent to the target node followed by the command which was redirected.
- [ ] CLUSTER ADDSLOTS slot [slot ...] Assign new hash slots to receiving node
- [ ] CLUSTER BUMPEPOCH Advance the cluster config epoch
- [ ] CLUSTER COUNT-FAILURE-REPORTS node-id Return the number of failure reports active for a given node
- [ ] CLUSTER COUNTKEYSINSLOT slot Return the number of local keys in the specified hash slot
- [ ] CLUSTER DELSLOTS slot [slot ...] Set hash slots as unbound in receiving node
- [ ] CLUSTER FAILOVER [FORCE|TAKEOVER] Forces a replica to perform a manual failover of its master.
- [ ] CLUSTER FLUSHSLOTS Delete a node's own slots information
- [ ] CLUSTER FORGET node-id Remove a node from the nodes table
- [ ] CLUSTER GETKEYSINSLOT slot count Return local key names in the specified hash slot
- [ ] CLUSTER INFO Provides info about Redis Cluster node state
- [ ] CLUSTER KEYSLOT key Returns the hash slot of the specified ke- [ ]
- [ ] CLUSTER MEET ip port Force a node cluster to handshake with another node
- [ ] CLUSTER MYID Return the node id
- [ ] CLUSTER NODES Get Cluster config for the node
- [ ] CLUSTER REPLICATE node-id Reconfigure a node as a replica of the specified master node
- [ ] CLUSTER RESET [HARD|SOFT] Reset a Redis Cluster node
- [ ] CLUSTER SAVECONFIG Forces the node to save cluster state on disk
- [ ] CLUSTER SET-CONFIG-EPOCH config-epoch Set the configuration epoch in a new node
- [x] CLUSTER SETSLOT slot IMPORTING|MIGRATING|STABLE|NODE [node-id] Bind a hash slot to a specific node
- [ ] CLUSTER SLAVES node-id List replica nodes of the specified master node
- [ ] CLUSTER REPLICAS node-id List replica nodes of the specified master node
- [x] CLUSTER SLOTS Get array of Cluster slot to node mappings- [ ]
- [ ] READONLY Enables read queries for a connection to a cluster replica node
- [ ] READWRITE Disables read queries for a connection to a cluster replica node

### Tips

anatolysergeev:
Partially done #151


testcontainers/redis-cluster-compose.yml
```diff
@@ -0,0 +1,17 @@
+version: '3.3'
+services:
+  redis-cluster:
+    image: 'redis:6.2.5-alpine'
+    container_name: redis
+    volumes:
+      - /home/alsergeev/workspace/zio/zio-redis/testcontainers/redis-cluster.sh:/data/redis-cluster.sh
```
mijicd, line 7: I don't think this is going to work :). Also, you may use something like this without the need to split the single compose file, and introduce custom scripts:

```yaml
  redis-node-0:
    image: docker.io/bitnami/redis-cluster:6.2
    environment:
      - 'ALLOW_EMPTY_PASSWORD=yes'
      - 'REDIS_NODES=redis-node-0 redis-node-1 redis-node-2 redis-node-3 redis-node-4 redis-node-5'

  redis-node-1:
    image: docker.io/bitnami/redis-cluster:6.2
    environment:
      - 'ALLOW_EMPTY_PASSWORD=yes'
      - 'REDIS_NODES=redis-node-0 redis-node-1 redis-node-2 redis-node-3 redis-node-4 redis-node-5'

  redis-node-2:
    image: docker.io/bitnami/redis-cluster:6.2
    environment:
      - 'ALLOW_EMPTY_PASSWORD=yes'
      - 'REDIS_NODES=redis-node-0 redis-node-1 redis-node-2 redis-node-3 redis-node-4 redis-node-5'

  redis-node-3:
    image: docker.io/bitnami/redis-cluster:6.2
    environment:
      - 'ALLOW_EMPTY_PASSWORD=yes'
      - 'REDIS_NODES=redis-node-0 redis-node-1 redis-node-2 redis-node-3 redis-node-4 redis-node-5'

  redis-node-4:
    image: docker.io/bitnami/redis-cluster:6.2
    environment:
      - 'ALLOW_EMPTY_PASSWORD=yes'
      - 'REDIS_NODES=redis-node-0 redis-node-1 redis-node-2 redis-node-3 redis-node-4 redis-node-5'

  redis-node-5:
    image: docker.io/bitnami/redis-cluster:6.2
    depends_on:
      - redis-node-0
      - redis-node-1
      - redis-node-2
      - redis-node-3
      - redis-node-4
    environment:
      - 'ALLOW_EMPTY_PASSWORD=yes'
      - 'REDIS_CLUSTER_REPLICAS=1'
      - 'REDIS_NODES=redis-node-0 redis-node-1 redis-node-2 redis-node-3 redis-node-4 redis-node-5'
      - 'REDIS_CLUSTER_CREATOR=yes'
    ports:
      - "6379:6379"
 ```

File: redis/src/test/scala/zio/redis/executor/cluster/RedisClusterExecutorLiveSpec.scala [hunk 2]
```diff
@@ -0,0 +1,79 @@
...
+    val executor = ZLayer.succeed(uri) >>> RedisExecutor.layer
+    (executor ++ ZLayer.succeed[Codec](ProtobufCodec) >>> RedisLive.layer).fresh
+  }
+
+  private val ClusterLayer = {
+    val address1      = RedisUri("localhost", 5010)
+    val address2      = RedisUri("localhost", 5000)
+    val clusterConfig = ZLayer.succeed(RedisClusterConfig(Chunk(address1, address2)))
+    val executor      = clusterConfig >>> RedisClusterExecutorLive.layer.orDie
+    val redis         = executor ++ ZLayer.succeed(codec) >>> RedisLive.layer
```
mijicd, line 76: Let's use `ZLayer.make` instead.

redis/src/test/scala/zio/redis/executor/cluster/RedisClusterExecutorLiveSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+          value3 <- get(key).returning[String] // have to get value without refreshing connection
+        } yield {
+          assertTrue(value1 == value2) && assertTrue(value2 == value3)
+        }
+      }
+    ).provideCustomLayerShared(ClusterLayer)
+
+  private final def getRedisNodeLayer(uri: RedisUri): Layer[Any, RedisEnv] = {
+    val executor = ZLayer.succeed(uri) >>> RedisExecutor.layer
+    (executor ++ ZLayer.succeed[Codec](ProtobufCodec) >>> RedisLive.layer).fresh
```
mijicd, line 68: Let's use `ZLayer.make` instead.

build.sbt
```diff
@@ -64,7 +64,8 @@ lazy val benchmarks =
         "dev.profunktor"    %% "redis4cats-effects"  % "1.2.0",
         "io.chrisdavenport" %% "rediculous"          % "0.4.0",
         "io.laserdisc"      %% "laserdisc-fs2"       % "0.5.0",
-        "dev.zio"           %% "zio-schema-protobuf" % "0.2.1"
+        "dev.zio"           %% "zio-schema-protobuf" % "0.2.1",
+        "org.redisson"       % "redisson"            % "3.16.8"
```
mijicd, line 68: I would not introduce it here. If you want to compare with redisson (which is a debatable idea), I'd do it in a follow-up PR.

redis/src/main/scala/zio/redis/RedisCommand.scala
```diff
@@ -18,18 +18,21 @@ package zio.redis
...
-      .serviceWithZIO[Redis] { redis =>
-        val command = Varargs(StringInput).encode(name.split(" "))(redis.codec) ++ input.encode(in)(redis.codec)
+  private[redis] def run(in: In): ZIO[RedisEnv, RedisError, Out] = {
+    val zio =
+      for {
+        redis <- ZIO.service[Redis]
+        out   <- redis.executor.execute(resp(in, redis.codec))
+        res   <- ZIO.attempt(output.unsafeDecode(out)(redis.codec))
+      } yield res
+    zio.refineToOrDie[RedisError]
```
mijicd, line 31: Was this necessary?

redis/src/main/scala/zio/redis/Output.scala
```diff
@@ -25,22 +26,8 @@ sealed trait Output[+A] {
...
-      case RespValue.Error(msg) if msg.startsWith("NOSCRIPT") =>
-        throw RedisError.NoScript(msg.drop(8).trim)
-      case RespValue.Error(msg) if msg.startsWith("NOTBUSY") =>
-        throw RedisError.NotBusy(msg.drop(7).trim)
-      case RespValue.Error(msg) =>
-        throw RedisError.ProtocolError(msg.trim)
-      case success =>
-        tryDecode(success)
+      case error: RespValue.Error => throw error.toRedisError
+      case success                => tryDecode(success)
```
mijicd, line 30: Was this necessary?

redis/src/main/scala/zio/redis/RedisUri.scala
```diff
@@ -16,8 +16,17 @@
 
 package zio.redis
 
-final case class RedisConfig(host: String, port: Int)
+final case class RedisUri(host: String, port: Int) {
```
mijicd, line 19: Please refrain from making unrelated changes. Regarding this particular case, we have already discussed it in the past, so please leave it as it is.

redis/src/main/scala/zio/redis/RedisUri.scala
```diff
@@ -16,8 +16,17 @@
...
+object RedisUri {
+  lazy val Default: RedisUri = RedisUri("localhost", 6379)
 
-object RedisConfig {
-  lazy val Default: RedisConfig = RedisConfig("localhost", 6379)
+  def apply(hostAndPort: String): RedisUri = {
+    val splitting = hostAndPort.split(':')
+    val host      = splitting(0)
+    val port      = splitting(1).toInt
+    RedisUri(host, port)
```
mijicd, line 30: Why is this necessary?

redis/src/main/scala/zio/redis/api/Cluster.scala
```diff
@@ -0,0 +1,76 @@
...
+
+trait Cluster {
+
+  def asking(): ZIO[RedisEnv, RedisError, Unit] =
+    AskingCommand.run(())
+
+  def slots(): ZIO[RedisEnv, RedisError, Chunk[Partition]] = {
+    val command = RedisCommand(ClusterSlots, NoInput, ChunkOutput(ClusterPartitionOutput))
+    command.run(())
+  }
```
mijicd, line 35: I would drop the parentheses.

redis/src/main/scala/zio/redis/ResultBuilder.scala
```diff
@@ -33,18 +33,18 @@ object ResultBuilder {
   final abstract class NeedsReturnType
 
   trait ResultBuilder1[+F[_]] extends ResultBuilder {
-    def returning[R: Schema]: ZIO[Redis, RedisError, F[R]]
+    def returning[R: Schema]: ZIO[RedisEnv, RedisError, F[R]]
```
mijicd, line 36: Please refrain from making non-related changes. Renaming `Redis` to `RedisEnv` does not bring any value other than increasing the diff size and introducing a confusing name, so please restore it.

redis/src/main/scala/zio/redis/codec/StringUtf8Codec.scala
```diff
@@ -0,0 +1,63 @@
...
+import zio.redis.RedisError.CodecError
+import zio.schema.Schema
+import zio.schema.StandardType.{DoubleType, IntType, LongType}
+import zio.schema.codec.Codec
+import zio.stream.ZPipeline
+import zio.{Chunk, ZIO}
+
+import java.nio.charset.StandardCharsets
+
+private[redis] object StringUtf8Codec extends Codec {
```
mijicd, line 27: Why is it restored?

redis/src/main/scala/zio/redis/executor/RedisConnection.scala
```diff
@@ -0,0 +1,135 @@
...
+import zio._
+import zio.redis.{RedisError, RedisUri, logScopeFinalizer}
+import zio.stream.{Stream, ZStream}
+
+import java.io.{EOFException, IOException}
+import java.net.{InetSocketAddress, SocketAddress, StandardSocketOptions}
+import java.nio.ByteBuffer
+import java.nio.channels.{AsynchronousSocketChannel, Channel, CompletionHandler}
+
+private[redis] trait RedisConnection {
```
mijicd, line 28: If I am not mistaken, this was formerly known as `ByteStream`? While I think that the name change is good, there are two problems associated with it:
- package hierarchy restructuring
- rename diff

I would ask you to refrain from making non-related changes and restore the previous structure (you can keep the new name).

redis/src/main/scala/zio/redis/executor/RedisConnection.scala
```diff
@@ -0,0 +1,135 @@
...
+  def read: Stream[IOException, Byte]
+  def write(chunk: Chunk[Byte]): IO[IOException, Option[Unit]]
+}
+
+private[redis] final class RedisConnectionLive(
```
mijicd, line 33: Let's keep the structure that was present before, i.e. one top-level artifact per file.

redis/src/main/scala/zio/redis/executor/RedisExecutor.scala
```diff
@@ -0,0 +1,21 @@
+package zio.redis.executor
```
mijicd, line 1: Same thing.

redis/src/main/scala/zio/redis/executor/cluster/RedisClusterConfig.scala
```diff
@@ -0,0 +1,6 @@
+package zio.redis.executor.cluster
```
mijicd, line 1: I would not nest packages this much.

redis/src/main/scala/zio/redis/executor/cluster/RedisClusterExecutorLive.scala
```diff
@@ -0,0 +1,169 @@
...
+        case success            => ZIO.succeed(success)
+      }.catchSome {
+        case e: RedisError.Ask   => executeAsk(getExecutor(e.address), asking = true)
+        case _: RedisError.Moved => refreshConnect() *> executeAsk(getExecutor(keySlot))
+      }
+      recover.retry(Scheduler)
+    }
+
+    for {
+      key    <- Try(command(1)).fold(_ => ZIO.fail(CusterKeyError), bs => ZIO.succeedNow(bs))
```
mijicd, line 58: Why not `ZIO.attempt`?

mijicd (commented):
Thanks for taking care of this!

testcontainers/redis-cluster-compose.yml
anatolysergeev: Ok. I spent half of a day trying your suggestion. It doesn't work or i don't know how to do it work(( I can't change ports in cluster to expose them, for redis cluster ports that expose have to equals node port. I suppose bitnami has a bug in scripts that starting cluster.
My script is simple from redis documentation, i suggest stick to it.

build.sbt
anatolysergeev: Sorry. I forgot to delete it. Just use it to explore redisson implementation of cluster.

testcontainers/redis-cluster-compose.yml
mijicd: I literally copied this setup from a working test setup I have, so, yes, it works.

mijicd: You can leave the script, but please don't restructure the directories.

anatolysergeev: Have u run tests? What works?

mijicd: Let's go one more time: the script i pasted is copied from a working test setup I have and I use at work, so yes, I am running tests against it :).

Nevertheless, to end this discussion as it is already going the wrong way, as I said in the comment above: please keep the script you added and don't restructure the compose files .

anatolysergeev: Sorry, if u understand me wrong. I'm just trying to figure out, what i'm doing wrong, cause for me it doesn't work.
It's starting cluster - yes. But i can't connect to it cause our tests not in the same network that this containers. That's why i have to expose ports, but if exposing ports do not equal redis node ports, redis cluster config know nothing about expose ports and our client would try to connect to redis node ports inside container.

redis/src/main/scala/zio/redis/RedisCommand.scala
anatolysergeev: Don't understand what was necessary?
if u ask about put encode logic in separate method - yes. Cause I'm using it in redis cluster ask command. Don't know how to do it better with current architecture

testcontainers/redis-cluster-compose.yml
anatolysergeev: OK. Yeah my local filesystem references i already deleted)

anatolysergeev: > You can leave the script, but please don't restructure the directories.

Are you asking to delete testcontainers folder?

redis/src/main/scala/zio/redis/RedisCommand.scala
mijicd: What i see here is that you replaced the serviceWithZIO approach with an equivalent code written via for comprehension which is not really related to the PR scope. Please correct me if I missed something.

testcontainers/redis-cluster-compose.yml
mijicd: Yes.

anatolysergeev: Isn't it better to have every thing tidy and not in one place?
I like everything in it's folders and with logic

mijicd: Ok.

mijicd: > OK. Yeah my local filesystem references i already deleted) 

I presume you didn't push that yet cause i still see it :)

redis/src/main/scala/zio/redis/RedisCommand.scala
anatolysergeev: The main point was to place logic ```Varargs(StringInput).encode(name.split(" "))(codec) ++ input.encode(in)(codec)``` to separate method to call it from RedisClusterExecutorLive
The refactoring was just cause i thought it's new better approach if you want i cat return serviceWithZIO

redis/src/test/scala/zio/redis/executor/cluster/RedisClusterExecutorLiveSpec.scala
anatolysergeev: done

redis/src/test/scala/zio/redis/executor/cluster/RedisClusterExecutorLiveSpec.scala
anatolysergeev: done

redis/src/main/scala/zio/redis/RedisCommand.scala
mijicd: Factoring the encoding out is perfectly fine, the rest is not necessary so please revert it.

anatolysergeev: done

redis/src/main/scala/zio/redis/Output.scala
anatolysergeev: I need this logic in RedisClusterExecutorLive to know what error returned

redis/src/main/scala/zio/redis/RedisUri.scala
anatolysergeev: I think that bad idea to have in RedisConfig class just host and port variables. First of all i need this structure in RedisClusterConfig and that would be strange imho RedisClusterConfig.config: RedisConfig. All libraries have class for node address.

redis/src/main/scala/zio/redis/RedisUri.scala
anatolysergeev: Cause slot errors returns in this format where i should redirect my request

redis/src/main/scala/zio/redis/RedisUri.scala
mijicd: We have already discussed this and agreed not to change it. The fact that it now contains only host and port is very likely temporary. Please do not change things that are already discussed and not related to the scope of a particular pull request.

redis/src/main/scala/zio/redis/api/Cluster.scala
anatolysergeev: How?) it has two lines

redis/src/main/scala/zio/redis/ResultBuilder.scala
anatolysergeev: I added when we had zio 1, and i have to propagate everywhere logger.
No it useless - return Redis)

redis/src/main/scala/zio/redis/codec/StringUtf8Codec.scala
anatolysergeev: cause i need light codec to execute commands in RedisClusterExecutorLive before connecting to cluster (request cluster configs, request ask command)

redis/src/main/scala/zio/redis/executor/cluster/RedisClusterExecutorLive.scala
anatolysergeev: Cause it can't fail) Why 'ZIO.attempt' ?)

redis/src/main/scala/zio/redis/executor/RedisConnection.scala
anatolysergeev: I think it's good time to restructure package hierarchy and rename classes other way with cluster class it would be full mess. It's better to do now in first steps. And think a little bit further about where structure would develop.

redis/src/main/scala/zio/redis/RedisUri.scala
anatolysergeev: What are u suggesting i need to create RedisUri separate from redis config to use it in RedisClusterConfig or to use inside RedisConfig

testcontainers/redis-cluster-compose.yml
anatolysergeev: Interesting. May this is because im force pushed some commits 🤔

redis/src/main/scala/zio/redis/api/Cluster.scala
mijicd: Parentheses, not braces, e.g.:

```
def scope:
```

redis/src/main/scala/zio/redis/executor/cluster/RedisClusterExecutorLive.scala
mijicd: I am suggesting that you drop the Try and go through ZIO directly, and you are replying that it can't fail. If it can't fail, why are you using Try at all? :)

redis/src/main/scala/zio/redis/executor/RedisConnection.scala
mijicd: What is messy? Top-level package has ~10 artifacts. Please don't pollute a very worthy pull request with package reorganization that is not necessary.

redis/src/main/scala/zio/redis/executor/cluster/RedisClusterExecutorLive.scala
anatolysergeev: Oh. I'm so stupid, sorry my bad. Don't understand that you are talking about Try, thanks. Ofcourse it is better to use attempt.

mijicd: All good :)

redis/src/main/scala/zio/redis/api/Cluster.scala
anatolysergeev: Oh, I understand, Ok - fixed.
In my humble opinion it's better with parentheses, for methods with side effects. Like in best practice from scala doc
https://docs.scala-lang.org/scala3/book/methods-most.html

redis/src/main/scala/zio/redis/executor/RedisConnection.scala
anatolysergeev: Ok. If i revert this structure in current PR, can I open another PR with this reorganization of folders?

mijicd: I am against it in general case. We don't need to split a package with ~10 artifacts into 5 packages with 2. It just makes navigation more difficult, and doesn't bring any value. If it grows significantly, then I don't object doing it.

anatolysergeev: Ok. I guess we found main disagreement) It's ok to have disagreements about code structure.))
If you find this structure is a bad idea of course it should be reverted and would be done when we find agreement)

In my opinion it is opposite. A proper structure help navigate through project and make shape of it, that new people can easily find that they want and understand where to put stuff. 

Reverted

anatolysergeev:
@mijicd Help) Can u suggest why after reverting some changes, compilation for scala 3.2.0 is failing? 😇

redis/src/main/scala/zio/redis/RedisUri.scala
anatolysergeev: Returned RedisConfig

anatolysergeev:
> @mijicd Help) Can u suggest why after reverting some changes, compilation for scala 3.2.0 is failing? innocent

fixed

redis/src/main/scala/zio/redis/RedisClusterConfig.scala
```diff
@@ -0,0 +1,18 @@
+package zio.redis
+
+import zio.Chunk
+
+final case class RedisUri(host: String, port: Int) {
+  override def toString: String = s"$host:$port"
+}
+
```
mijicd, line 8: Please keep a single top-level artifact per file. Also, please revert renaming `redis.scala` for the same reason.

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
```diff
@@ -0,0 +1,165 @@
...
+        _       <- ZIO.foreachParDiscard(connection.executors) { case (_, es) => es.scope.close(Exit.unit) }
+      } yield cluster
+    }
+
+  // TODO configurable
+  private val Scheduler: Schedule[Any, Throwable, (zio.Duration, Long, Throwable)] =
+    Schedule.exponential(100.millis, 1.5) && Schedule.recurs(5) && Schedule.recurWhile[Throwable] {
+      case _: RedisError.IOError | _: RedisError.ClusterRedisError => true
+      case _                                                       => false
+    }
```
mijicd, line 85: It can be moved to the companion object. Also, it's not a scheduler, rather a retry policy.

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
```diff
@@ -0,0 +1,165 @@
...
+
+  // TODO introduce max connection amount
+  private def getExecutor(address: RedisUri) =
+    clusterConnectionRef.modifyZIO { cc =>
+      val executorOpt       = cc.executors.get(address).map(es => (es.executor, cc))
+      val enrichedClusterIO = scope.extend(connectToNode(address)).map(es => (es.executor, cc.addExecutor(address, es)))
+      ZIO.fromOption(executorOpt).catchAll(_ => enrichedClusterIO)
+    }
+
+  private def refreshConnect(): ZIO[Any, RedisError, Unit] =
```
mijicd, line 71:
```suggestion
  private def refreshConnect: IO[RedisError, Unit] =
```

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
```diff
@@ -0,0 +1,165 @@
...
+      keySlot = Slot((key.asCRC16 % SlotsAmount).toLong)
+      result <- executeSafe(keySlot)
+    } yield result
+  }
+
+  private def getExecutor(slot: Slot): UIO[RedisExecutor] =
+    clusterConnectionRef.get.map(cc => cc.executor(slot))
+
+  // TODO introduce max connection amount
+  private def getExecutor(address: RedisUri) =
```
mijicd, line 64: Please add return type to be consistent.

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
```diff
@@ -0,0 +1,165 @@
...
+        executor <- executorZIO
+        _        <- ZIO.when(asking)(executor.execute(AskingCommand.resp((), StringUtf8Codec)))
+        res      <- executor.execute(command)
+      } yield res
+
+    def executeSafe(keySlot: Slot) = {
+      val recover = executeAsk(getExecutor(keySlot)).flatMap {
+        case e: RespValue.Error => ZIO.fail(e.toRedisError)
+        case success            => ZIO.succeed(success)
+      }.catchSome {
```
mijicd, line 46: No need for `flatMap` here, it can be `map`.

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
```diff
@@ -0,0 +1,165 @@
...
+) extends RedisExecutor {
+
+  def execute(command: Chunk[RespValue.BulkString]): IO[RedisError, RespValue] = {
+
+    def executeAsk(executorZIO: IO[RedisError.IOError, RedisExecutor], asking: Boolean = false) =
+      for {
+        executor <- executorZIO
+        _        <- ZIO.when(asking)(executor.execute(AskingCommand.resp((), StringUtf8Codec)))
+        res      <- executor.execute(command)
+      } yield res
```
mijicd, line 40: I would split this function.

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
```diff
@@ -0,0 +1,165 @@
...
+
+    for {
+      key    <- ZIO.attempt(command(1)).orElseFail(CusterKeyError)
+      keySlot = Slot((key.asCRC16 % SlotsAmount).toLong)
+      result <- executeSafe(keySlot)
+    } yield result
+  }
+
+  private def getExecutor(slot: Slot): UIO[RedisExecutor] =
+    clusterConnectionRef.get.map(cc => cc.executor(slot))
```
mijicd, line 61:
```suggestion
    clusterConnectionRef.get.map(_.executor(slot))
```

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
```diff
@@ -0,0 +1,165 @@
...
+
+object RedisClusterExecutorLive {
+
+  lazy val layer: ZLayer[RedisClusterConfig, RedisError, RedisExecutor] = ZLayer.scoped {
+    for {
+      config       <- ZIO.service[RedisClusterConfig]
+      layerScope   <- ZIO.scope
+      clusterScope <- Scope.make
+      executor     <- clusterScope.extend(create(config, clusterScope))
+      _            <- layerScope.addFinalizerExit(e => clusterScope.close(e))
```
mijicd, line 97:
```suggestion
      _            <- layerScope.addFinalizerExit(clusterScope.close)
```

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
```diff
@@ -0,0 +1,165 @@
...
+    Schedule.exponential(100.millis, 1.5) && Schedule.recurs(5) && Schedule.recurWhile[Throwable] {
+      case _: RedisError.IOError | _: RedisError.ClusterRedisError => true
+      case _                                                       => false
+    }
+
+}
+
+object RedisClusterExecutorLive {
+
+  lazy val layer: ZLayer[RedisClusterConfig, RedisError, RedisExecutor] = ZLayer.scoped {
```
mijicd, line 91: Super nit-pick: let's move `ZLayer.scoped` to the next line.

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
```diff
@@ -0,0 +1,165 @@
...
+    } yield clusterExec
+
+  private def initConnectToCluster(
+    addresses: Chunk[RedisUri]
+  ): ZIO[Scope, RedisError, ClusterConnection] =
+    ZIO
+      .collectFirst(addresses) { address =>
+        connectToCluster(address).foldZIO(
+          error => ZIO.logError(s"The connection to cluster has been failed, $error") *> ZIO.succeedNow(None),
+          cc => ZIO.logInfo("The connection to cluster has been established") *> ZIO.succeedNow(Some(cc))
```
mijicd, line 119: You can use `as` instead of lifting to ZIO and using `*>`.

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
```diff
@@ -0,0 +1,165 @@
...
+
+  private def getSlotAddress(partitions: Chunk[Partition]) =
+    partitions.flatMap { p =>
+      for (i <- p.slotRange.start to p.slotRange.end) yield Slot(i) -> p.master.address
+    }.toMap
+
+  private val CusterKeyError =
+    RedisError.ProtocolError("Key doesn't found. No way to dispatch this command to Redis Cluster")
+  private val CusterConnectionError =
+    RedisError.IOError(new IOException("The connection to cluster has been failed. Can't reach a single startup node."))
```
mijicd, line 164: Let's make both of 'em final.

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
```diff
@@ -0,0 +1,165 @@
...
+  private def connectToNode(address: RedisUri) =
+    for {
+      closableScope <- Scope.make
+      connection    <- closableScope.extend(RedisConnectionLive.create(RedisConfig(address.host, address.port)))
+      executor      <- closableScope.extend(RedisNodeExecutorLive.create(connection))
+      layerScope    <- ZIO.scope
+      _             <- layerScope.addFinalizerExit(e => closableScope.close(e))
+    } yield ExecutorScope(executor, closableScope)
+
+  private def getRedis(address: RedisUri) = {
```
mijicd, line 145: Let's drop `get` prefixes everywhere.

redis/src/main/scala/zio/redis/RedisNodeExecutorLive.scala
```diff
@@ -0,0 +1,102 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.redis
+
+import zio._
+import zio.redis.RedisNodeExecutorLive._
+
+final class RedisNodeExecutorLive(
```
mijicd, line 22: Let's rename all of them:
- `RedisClusterExecutorLive` to `ClusterExecutor`
- `RedisNodeExecutorLive` to `SingleNodeExecutor`
- `RedisTestExecutorLive` to `TestExecutor`

To be completely "ZIO2 friendly" we should make them publicly accessible and move all of the "live" layers into their companion objects.

redis/src/main/scala/zio/redis/api/Streams.scala
```diff
@@ -759,7 +759,7 @@ trait Streams {
   }
 }
 
-private object Streams {
+private[redis] object Streams {
```
mijicd, line 762: Why is this necessary?

redis/src/main/scala/zio/redis/codec/CRC16.scala
```diff
@@ -0,0 +1,42 @@
...
+ *   - Name: "XMODEM", also known as "ZMODEM", "CRC-16/ACORN"
+ *   - Poly: 1021 (That is actually ```x^16 + x^12 + x^5 + 1```)
+ *   - Initialization: 0000
+ *   - Reflect Input byte: False
+ *   - Reflect Output CRC: False
+ *   - Xor constant to output CRC: 0000
+ *   - Output for "123456789": 31C3
+ */
+private[redis] object CRC16 {
+  private val Lookup = Array(0x0000, 0x1021, 0x2042, 0x3063, 0x4084, 0x50a5, 0x60c6, 0x70e7, 0x8108, 0x9129, 0xa14a,
```
mijicd, line 19: Let's make it final and move `Array` to the new line.

redis/src/main/scala/zio/redis/options/Cluster.scala
```diff
@@ -0,0 +1,69 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package zio.redis.options
+
+import zio.redis.{RedisExecutor, RedisUri}
+import zio.{Chunk, Scope}
+
+trait Cluster {}
```
mijicd, line 21: Is this marker necessary?

redis/src/main/scala/zio/redis/options/Cluster.scala
```diff
@@ -0,0 +1,69 @@
...
+  case class SlotRange(start: Long, end: Long) {
+    def isContain(slot: Slot): Boolean = start <= slot.number && slot.number <= end
+  }
+
+  case class Partition(slotRange: SlotRange, master: Node, slaves: Chunk[Node]) {
+    lazy val nodes: Chunk[Node]         = master +: slaves
+    lazy val addresses: Chunk[RedisUri] = nodes.map(_.address)
+  }
+
+  sealed trait SetSlotSubCommand extends Product { self =>
```
mijicd, line 53: Let's make them final.

redis/src/main/scala/zio/redis/options/Cluster.scala
```diff
@@ -0,0 +1,69 @@
...
+  private[redis] final val SlotsAmount = 16384
+
+  case class ExecutorScope(executor: RedisExecutor, scope: Scope.Closeable)
+
+  case class ClusterConnection(
+    partitions: Chunk[Partition],
+    executors: Map[RedisUri, ExecutorScope],
+    slots: Map[Slot, RedisUri]
+  ) {
+    def executor(slot: Slot): RedisExecutor = executors(slots(slot)).executor
```
mijicd, line 34: This is an unsafe call.

redis/src/main/scala/zio/redis/options/Cluster.scala
```diff
@@ -0,0 +1,69 @@
...
+    def addExecutor(uri: RedisUri, es: ExecutorScope): ClusterConnection =
+      this.copy(executors = executors + (uri -> es))
+  }
+
+  case class Slot(val number: Long) extends AnyVal
+
+  case class Node(id: String, address: RedisUri)
+
+  case class SlotRange(start: Long, end: Long) {
+    def isContain(slot: Slot): Boolean = start <= slot.number && slot.number <= end
```
mijicd, line 45: Pleases rename to `contains`

redis/src/main/scala/zio/redis/options/Cluster.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  case class ClusterConnection(
+    partitions: Chunk[Partition],
+    executors: Map[RedisUri, ExecutorScope],
+    slots: Map[Slot, RedisUri]
+  ) {
+    def executor(slot: Slot): RedisExecutor = executors(slots(slot)).executor
+
+    def addExecutor(uri: RedisUri, es: ExecutorScope): ClusterConnection =
+      this.copy(executors = executors + (uri -> es))
```
mijicd, line 37: You can drop this.

redis/src/main/scala/zio/redis/package.scala
```diff
@@ -16,6 +16,8 @@
 
 package zio
 
+import zio.Exit.{Failure, Success}
```
mijicd, line 19: I'd drop this and qualify them on call site.

redis/src/main/scala/zio/redis/RedisConnection.scala
```diff
@@ -0,0 +1,134 @@
...
+
+import zio._
+import zio.stream.{Stream, ZStream}
+
+import java.io.{EOFException, IOException}
+import java.net.{InetSocketAddress, SocketAddress, StandardSocketOptions}
+import java.nio.ByteBuffer
+import java.nio.channels.{AsynchronousSocketChannel, Channel, CompletionHandler}
+
+private[redis] trait RedisConnection {
```
mijicd, line 27: Let's keep one top-level artifact per file.

mijicd (commented):
2nd pass done

redis/src/main/scala/zio/redis/RedisClusterConfig.scala
anatolysergeev: >Also, please revert renaming redis.scala for the same reason.

Sry, I don't understand what i should revert

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
anatolysergeev: I want to do it configurable, if i do it, should i keep it here?

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
anatolysergeev: Sry. I don't understand how it can be map, if i need zio to fail

redis/src/main/scala/zio/redis/RedisClusterConfig.scala
mijicd: There was a file named redis.scala which you renamed to Redis.scala.

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
mijicd: You are right, misread it.

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
mijicd: Sure, if you plan to make it flexible keep it here.

redis/src/main/scala/zio/redis/RedisClusterConfig.scala
anatolysergeev: Oh. Ok, done. I thought it was spelling mistake )

redis/src/main/scala/zio/redis/options/Cluster.scala
anatolysergeev: No, Just for consistency with other) 
Drop it.

redis/src/main/scala/zio/redis/api/Streams.scala
anatolysergeev: Unnecessary. Just for consistency.
revert it

redis/src/main/scala/zio/redis/RedisNodeExecutorLive.scala
anatolysergeev: I hope, that i did everything correct

redis/src/main/scala/zio/redis/RedisClusterExecutorLive.scala
anatolysergeev: Done

redis/src/main/scala/zio/redis/api/Cluster.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def setSlotStable(slot: Slot): ZIO[Redis, RedisError, Unit] = {
+    val command = RedisCommand(ClusterSetSlots, Tuple2(LongInput, ArbitraryInput[String]()), UnitOutput)
+    command.run((slot.number, Stable.stringify))
+  }
+
+  def setSlotMigrating(slot: Slot, destinationNodeId: String): ZIO[Redis, RedisError, Unit] = {
```
mijicd, line 42: Super nitpicky: I'd rename this to nodeId.

redis/src/main/scala/zio/redis/api/Cluster.scala
```diff
@@ -0,0 +1,76 @@
...
+    val command = RedisCommand(
+      ClusterSetSlots,
+      Tuple3(LongInput, ArbitraryInput[String](), ArbitraryInput[String]()),
+      UnitOutput
+    )
+    command.run((slot.number, Migrating.stringify, destinationNodeId))
+  }
+
+  def setSlotImporting(slot: Slot, sourceNodeId: String): ZIO[Redis, RedisError, Unit] = {
```
mijicd, line 51: Same here.

redis/src/main/scala/zio/redis/api/Cluster.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def asking: ZIO[Redis, RedisError, Unit] =
+    AskingCommand.run(())
+
+  def slots: ZIO[Redis, RedisError, Chunk[Partition]] = {
+    val command = RedisCommand(ClusterSlots, NoInput, ChunkOutput(ClusterPartitionOutput))
+    command.run(())
+  }
+
+  def setSlotStable(slot: Slot): ZIO[Redis, RedisError, Unit] = {
```
mijicd, line 37: Not sure if we should have these as separate functions, but if I'm not mistaken, we're already doing that elsewhere, so it's consistent :+1:

redis/src/main/scala/zio/redis/codec/CRC16.scala
```diff
@@ -0,0 +1,43 @@
+package zio.redis.codec
+
+import zio.Chunk
+
+/**
```
mijicd, line 5: Super nitpicky:
```suggestion
/*
```

redis/src/main/scala/zio/redis/options/Cluster.scala
```diff
@@ -0,0 +1,67 @@
...
+    executors: Map[RedisUri, ExecutorScope],
+    slots: Map[Slot, RedisUri]
+  ) {
+    def executor(slot: Slot): Option[RedisExecutor] = executors.get(slots(slot)).map(_.executor)
+
+    def addExecutor(uri: RedisUri, es: ExecutorScope): ClusterConnection =
+      copy(executors = executors + (uri -> es))
+  }
+
+  final case class Slot(val number: Long) extends AnyVal
```
mijicd, line 38: You should be able to drop the `val`.

redis/src/main/scala/zio/redis/options/Cluster.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  final case class Node(id: String, address: RedisUri)
+
+  final case class SlotRange(start: Long, end: Long) {
+    def contains(slot: Slot): Boolean = start <= slot.number && slot.number <= end
+  }
```
mijicd, line 44: Is it feasible to reuse range?

mijicd (approved):
A few minor remarks from my side, otherwise looks good to go. Before merging it, please adjust the PR comment to avoid accidentally closing the issue (i.e., right now, it contains a "magic" word `close`), and update the list in the issue.

Thank you!

redis/src/main/scala/zio/redis/options/Cluster.scala
anatolysergeev: Sry don't understand (( what range?

mijicd: Scala's.

redis/src/main/scala/zio/redis/codec/CRC16.scala
anatolysergeev: But, we everywhere in documentation use 2 stars. It's necessary to be green and formaIted like documentation in IntelIJ Idea)

mijicd: For public scaladoc, yes. However, this thing is (package) private and I'd argue that it should not be present there.

redis/src/main/scala/zio/redis/api/Cluster.scala
anatolysergeev: Yeah, we have  examples where method that can have different behavior depending on it's parameters in redis, we split by different methods in client

redis/src/main/scala/zio/redis/codec/CRC16.scala
anatolysergeev: In other redis libraries it's present, but ok i would rid of one star)

redis/src/main/scala/zio/redis/api/Cluster.scala
anatolysergeev: Done.
I remembered that forgot to add documentation and without it, renaming wasn't obvious what a node should be specified. 
Now with documentation it's clear.

mijicd (approved):

PR #655, opened 2022-10-31
Implement remaining script API commands

rfmejia:
Addresses remaining items in #158.

Please check if tests for these commands are satisfactory, I could only come up with a basic test for `scriptDebug`, and for `scriptKill` I couldn't figure out how to fork a long-running script (e.g., infinite loop) and attempt to kill it.


rfmejia:
Closed accidentally, please ignore

PR #711, opened 2023-01-01
Ignore Snapshot Versions on The README Page

khajavi:
There was a bug in previous versions of the zio-sbt-website plugin which makes an endless loop when the library hasn't been released yet.


khajavi:
There was a bug in previous versions of the zio-sbt-website plugin which makes an endless loop when the library hasn't been released yet.

mijicd (approved):

Issue #548, opened 2022-02-15
Introduce all Redis operations as type-safe methods of RedisExecutor trait


jdegoes:
In order to facilitate better use by layers, which want to consume other services by their interfaces, it would be ideal to add to `RedisExecutor` typed methods for all Redis operations, which simply encode/decode as appropriate.

This way, a service can depend on RedisExecutor, and use its operations, without also directly adding any dependencies into ZIO environment (which would have to be implemented for a service which is calling RedisExecutor).

anatolysergeev:
Recently, we have changed [service structure](https://github.com/zio/zio-redis/pull/552) and i suppose now it should be `Redis` trait. Am i right?

rfmejia:
Looking into this while migrating a codebase to ZIO 2, I have some questions:

1. Is this the desired result, moving the following from the `zio.redis` package object:

```scala
trait Redis extends api.Connection
  with api.Geo 
  with api.Hashes
  ...
{
  def codec: Codec
  def executor: RedisExecutor
}
```

2. Should accessor methods remain in the package object, or should they be moved to the `zio.redis.Redis` companion object?

3. Where should `options.XXX` be placed?

mijicd:
@rfmejia The idea is to move the api functions to `Redis` trait so that it "plays nice" with the rest of the ecosystem as @jdegoes described above. Eventually, we'll hide `codec` and `executor` methods, but I'd say it's not a concern of this particular issue. Helper packages, like options, should not be touched.

m-kalai:
i'm going to take this one but let me check if i understand what the end result should look like. 
Goal is to have `Redis` trait (and probably companion object as well) with all the `api` method declarations, and those declarations probably shouldn't "leak" dependencies i.e. something like
```
trait Redis {
  def auth(password: String): IO[RedisError, Unit]
  ...
}
```
which would require changes in api traits and the way commands are run. Or `Redis` trait could be just mixed in with `api` traits but all the declarations should change to not to "depend" on `Redis`.
Doesn't seem like a small or mundane task :) Or it could be i'm missing something, it's first time i'm looking at this codebase.

mijicd:
@m-kalai That is correct: a single "service" trait without dependency leaks. You are free to reuse / redefine / introduce traits if you think they'd improve the current state. The issue mentions RedisExecutor as it was the only trait at that time.

PR #725, opened 2023-01-12
[DRAFT] starting to move implementations from traits into Redis object

m-kalai:
Resolves https://github.com/zio/zio-redis/issues/548

To be able to use traits without "leaking" dependencies on `Redis` BUT also make sure that all these "contracts" have been fulfilled on the implementing `Redis` object I created this pattern where each API trait will have environment type `R`. 

This way we allow creation of in memory test implementation of a trait where `R` will be `Any` for example. But also we reuse same traits in `Redis` "accessor" object with R being `Redis`.

It's probably not a nicest solution but it's pretty flexible. Let me know if I can continue in this direction as this refactoring is quite time consuming. Other ideas are of course welcome.


mijicd:
@m-kalai methods should live on trait, not in its companion object. We can expand the companion with so called accessors at some point, but for now, the scope is exposing them via trait.

m-kalai:
> @m-kalai methods should live on trait, not in its companion object. We can expand the companion with so called accessors at some point, but for now, the scope is exposing them via trait.

you mean that implementation of those methods should remain on each API trait? then i need 2 traits per API, one where we don't leak dependency on Redis and another one with implementation of those commands (which depend on Redis). We can't do both in one trait unless we change api to return `RedisCommand` instead of `ZIO`.

mijicd:
No, I don't. See the issue description and the comments in there. The idea is to expose these methods through a single trait so that it plays nicely with the ["service pattern"](https://zio.dev/guides/migrate/zio-2.x-migration-guide/#service-pattern). 

Let's take `get` as an example. Currently, it lives in its commands trait and is available via the package object. However, the functions in all command traits depend on `Redis`, meaning you'll have to provide the executor whenever you invoke them, as in the following example:

```scala
trait SomeService {
  def readValue(key: String): UIO[String]
}

final class SomeServiceLive(redis: Redis) extends SomeService {
  def readValue(key: String) = zio.redis.get(key).provide(redis).orDie
}
```

If we reorganize it in a way that the complete functionality is available via the single trait, we will get something like this:

```scala
trait SomeService {
  def readValue(key: String): UIO[String]
}

final class SomeServiceLive(redis: Redis) extends SomeService {
  def readValue(key: String) = redis.get(key).orDie
}
```

m-kalai:
Seems to me like we aim to achieve the same thing, but for some reason I don't see how you can do it without doing any of those things i described.
To follow "service pattern" we need to have `Redis` trait which methods don't have any environment dependency. I think we established when i picked that task up. I'm gonna use `exists` in my examples instead of `get` which returns builder. 
So we need something like:
```scala
trait Redis {
  def exists[K: Schema](key: K, keys: K*): ZIO[Any, RedisError, Long]
}
```
If we have something like that then we can do something like you described above, i.e.
```scala
final class SomeServiceLive(redis: Redis) extends SomeService {
  def readValue(key: String) = redis.exists(key).orDie
}
```
but the `Keys` trait signature of this method is 
```scala
final def exists[K: Schema](key: K, keys: K*): ZIO[Redis, RedisError, Long]
```
which means that `Redis` trait can't extend `Keys` trait as the types are incompatible. But I want to have relationship between those traits otherwise what's the point. We might be missing methods in Redis trait and not even know about it. That's why i wanted to have base trait and something else implementing it (doesn't have to be the Redis companion object).

Other way to get `Redis` environment dependency would be something like 
```scala
trait Keys {
  def redis: Redis

  final def exists[K: Schema](key: K, keys: K*): ZIO[Any, RedisError, Long] = {
    redis.executor.run(command, (key, keys.toList))
  }
}
```
but that can bring its share of complications too. So what am i missing?

mijicd:
@m-kalai `Redis` trait is fully "aware" of all the dependencies necessary for the command to run (`codec` and `executor`).

m-kalai:
@mijicd yes `Redis` trait is, but implementing traits are not. I'd like to avoid just copying methods into `Redis` trait, there is 100+ of them so it would be quite error prone and we wouldn't have the guarantees that change in API reaches `Redis` trait.

Can I add `codec` and `executor` to each API trait? That would help me to get rid of environment dependency on `Redis` in all methods signatures and we could just mix in all the API traits in `Redis` trait.

mijicd:
I don't mind that, but make sure to remove the dependencies from those methods.

m-kalai:
cool, i'll keep this draft open for the approach no. 2

m-kalai:
i gave it a try, `Redis` trait now contains all the methods. What i haven't realised is that we're no longer able to create accessor object like `package object redis` was. At least not without breaking dependency injection layers offer. We'd have to manually create it by redefining all 100+ methods which to me sounds like a bad idea. If we had only a few methods like RocksDB has we probably could get away with it.

That's why i like having "base trait" with just interfaces/declarations and environment type R. Then i can use same trait for concrete implementation and accessor object. I'm using that lately with Scala 3 as there is no `zio-macros` support and `@accessible`.  By using same trait I don't have to worry about somebody making change on one place but not reflecting it on the other.

Not sure how to tackle this one as you didn't like my approach with base traits and i don't like copy pasting on such scale 😄

mijicd:
We no longer need the package object approach so feel free to nuke it :)

mijicd:
We no longer need the package object, so feel free to nuke it :)

m-kalai:
oh, i thought you want to keep it, or the version of it, as it's used in all the tests... ok if i can nuke it then it's just manual labour of fixing all the tests 🙂

PR #727, opened 2023-01-15
[548] Expose Redis operations via Redis trait

### LINKED ISSUE #548: Introduce all Redis operations as type-safe methods of RedisExecutor trait

m-kalai:
Resolves #548 

* getting rid of `Redis` environment dependency from all the API traits
* mixing API traits in the `Redis` trait
* getting rid of "accessors" from `redis` package object and fixing tests

### DIFF
--- a/example/src/main/scala/example/contributorsCache.scala
+++ b/example/src/main/scala/example/contributorsCache.scala
@@ -36,12 +36,15 @@ final case class ContributorsCacheLive(r: Redis, s: Sttp) extends ContributorsCa
     ZIO
       .fromOption(NonEmptyChunk.fromChunk(contributors))
       .map(Contributors(_).toJson)
-      .flatMap(data => set(repository.key, data, Some(1.minute)).orDie)
+      .flatMap(data => ZIO.serviceWithZIO[Redis](_.set(repository.key, data, Some(1.minute))).orDie)
       .ignore
 
   private def read(repository: Repository): ZIO[Redis, ApiError, Contributors] =
-    get(repository.key)
-      .returning[String]
+    ZIO
+      .serviceWithZIO[Redis](
+        _.get(repository.key)
+          .returning[String]
+      )
       .someOrFail(ApiError.CacheMiss(repository.key))
       .map(_.fromJson[Contributors])
       .foldZIO(_ => ZIO.fail(ApiError.CorruptedData), s => ZIO.succeed(s.getOrElse(Contributors(Chunk.empty))))
--- a/example/src/main/scala/example/contributorsCache.scala
+++ b/example/src/main/scala/example/contributorsCache.scala
@@ -41,10 +41,7 @@ final case class ContributorsCacheLive(r: Redis, s: Sttp) extends ContributorsCa
 
   private def read(repository: Repository): ZIO[Redis, ApiError, Contributors] =
     ZIO
-      .serviceWithZIO[Redis](
-        _.get(repository.key)
-          .returning[String]
-      )
+      .serviceWithZIO[Redis](_.get(repository.key).returning[String])
       .someOrFail(ApiError.CacheMiss(repository.key))
       .map(_.fromJson[Contributors])
       .foldZIO(_ => ZIO.fail(ApiError.CorruptedData), s => ZIO.succeed(s.getOrElse(Contributors(Chunk.empty))))
--- a/redis/src/main/scala/zio/redis/CommandExecutor.scala
+++ b/redis/src/main/scala/zio/redis/CommandExecutor.scala
@@ -1,9 +1,0 @@
-package zio.redis
-
-import zio.schema.codec.BinaryCodec
-
-trait CommandExecutor {
-  implicit def codec: BinaryCodec
-
-  implicit def executor: RedisExecutor
-}
--- a/redis/src/main/scala/zio/redis/CommandExecutor.scala
+++ b/redis/src/main/scala/zio/redis/CommandExecutor.scala
@@ -1,0 +1,9 @@
+package zio.redis
+
+import zio.schema.codec.BinaryCodec
+
+trait CommandExecutor {
+  implicit def codec: BinaryCodec
+
+  implicit def executor: RedisExecutor
+}
--- a/redis/src/main/scala/zio/redis/RedisCommand.scala
+++ b/redis/src/main/scala/zio/redis/RedisCommand.scala
@@ -30,6 +30,12 @@ final class RedisCommand[-In, +Out] private (val name: String, val input: Input[
       }
       .refineToOrDie[RedisError]
 
+  private[redis] def run(in: In)(implicit executor: RedisExecutor, codec: BinaryCodec): ZIO[Any, RedisError, Out] =
+    executor
+      .execute(resp(in, codec))
+      .flatMap[Any, Throwable, Out](out => ZIO.attempt(output.unsafeDecode(out)(codec)))
+      .refineToOrDie[RedisError]
+
   private[redis] def resp(in: In, codec: BinaryCodec): Chunk[RespValue.BulkString] =
     Varargs(StringInput).encode(name.split(" "))(codec) ++ input.encode(in)(codec)
 }
--- a/redis/src/main/scala/zio/redis/RedisCommand.scala
+++ b/redis/src/main/scala/zio/redis/RedisCommand.scala
@@ -20,27 +20,25 @@ import zio._
 import zio.redis.Input.{StringInput, Varargs}
 import zio.schema.codec.BinaryCodec
 
-final class RedisCommand[-In, +Out] private (val name: String, val input: Input[In], val output: Output[Out]) {
-  private[redis] def run(in: In): ZIO[Redis, RedisError, Out] =
-    ZIO
-      .serviceWithZIO[Redis] { redis =>
-        redis.executor
-          .execute(resp(in, redis.codec))
-          .flatMap[Any, Throwable, Out](out => ZIO.attempt(output.unsafeDecode(out)(redis.codec)))
-      }
-      .refineToOrDie[RedisError]
+final class RedisCommand[-In, +Out] private (val name: String, val input: Input[In], val output: Output[Out])(
+  val executor: RedisExecutor,
+  val codec: BinaryCodec
+) {
 
-  private[redis] def run(in: In)(implicit executor: RedisExecutor, codec: BinaryCodec): ZIO[Any, RedisError, Out] =
+  private[redis] def run(in: In): IO[RedisError, Out] =
     executor
-      .execute(resp(in, codec))
+      .execute(resp(in))
       .flatMap[Any, Throwable, Out](out => ZIO.attempt(output.unsafeDecode(out)(codec)))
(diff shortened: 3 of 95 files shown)


mijicd:
Am I missing something? This PR seems to contain only benchmarks.

m-kalai:
@mijicd you scared me there a bit :) but i see all the changes, not just in the benchmark module

mijicd:
> @mijicd you scared me there a bit :) but i see all the changes, not just in the benchmark module

I guess it's GitHub app issue. I'll check it out from laptop.

redis/src/main/scala/zio/redis/CommandExecutor.scala
```diff
@@ -0,0 +1,9 @@
+package zio.redis
+
+import zio.schema.codec.BinaryCodec
+
+trait CommandExecutor {
+  implicit def codec: BinaryCodec
+
+  implicit def executor: RedisExecutor
+}
```
mijicd, line 9: Let's rename it somehow, as this is not really an executor.

redis/src/main/scala/zio/redis/api/Lists.scala
```diff
@@ -20,10 +20,10 @@ import zio._
 import zio.redis.Input._
 import zio.redis.Output._
 import zio.redis.ResultBuilder._
-import zio.redis._
+import zio.redis.{CommandExecutor, _}
```
mijicd, line 23: This change can be reverted.

redis/src/main/scala/zio/redis/codecs/CRC16.scala
```diff
@@ -1,4 +1,4 @@
-package zio.redis.codec
+package zio.redis.codecs
```
mijicd, line 1: This change is not really related to the PR scope, but I like it so let it stay.

redis/src/test/scala/zio/redis/ClusterExecutorSpec.scala
```diff
@@ -9,53 +9,63 @@ object ClusterExecutorSpec extends BaseSpec {
...
-          _                = ZIO.logDebug(s"$key _____ Migrating $keySlot from ${sourceMaster.id}- ${sourceMaster.address}")
-          sourceMasterConn = getRedisNodeLayer(sourceMaster.address)
-          _               <- setSlotMigrating(keySlot, destMaster.id).provideLayer(sourceMasterConn)
-          value2          <- get(key).returning[String] // have to redirect without error ASK
-          value3          <- get(key).returning[String] // have to redirect without creating new connection
-          _               <- setSlotStable(keySlot).provideLayer(destMasterConn)
-        } yield {
-          assertTrue(value1 == value2) && assertTrue(value2 == value3)
+        ZIO.serviceWithZIO[Redis] { redis =>
+          import redis._
```
mijicd, line 13: Let's not do this.

redis/src/test/scala/zio/redis/ClusterExecutorSpec.scala
```diff
@@ -9,53 +9,63 @@ object ClusterExecutorSpec extends BaseSpec {
...
-          _ <- migrate(destMaster.address.host, destMaster.address.port.toLong, key, 0, 5.seconds, keys = None)
-                 .provideLayer(sourceMasterConn)
-          _      <- setSlotNode(keySlot, destMaster.id).provideLayer(destMasterConn)
-          _      <- setSlotNode(keySlot, destMaster.id).provideLayer(sourceMasterConn)
-          value2 <- get(key).returning[String] // have to refresh connection
-          value3 <- get(key).returning[String] // have to get value without refreshing connection
-        } yield {
-          assertTrue(value1 == value2) && assertTrue(value2 == value3)
+        ZIO.serviceWithZIO[Redis] { redis =>
+          import redis._
```
mijicd, line 39: Same here.

redis/src/test/scala/zio/redis/ConnectionSpec.scala
```diff
@@ -12,85 +12,97 @@ trait ConnectionSpec extends BaseSpec {
...
       suite("clientCaching")(
         test("track keys") {
-          for {
-            _            <- clientTrackingOff
-            _            <- clientTrackingOn(trackingMode = Some(ClientTrackingMode.OptIn))
-            _            <- clientCaching(true)
-            trackingInfo <- clientTrackingInfo
-          } yield assert(trackingInfo.flags.caching)(isSome(isTrue))
+          ZIO.serviceWithZIO[Redis] { redis =>
+            import redis._
```
mijicd, line 16: Same here.

redis/src/test/scala/zio/redis/ConnectionSpec.scala
```diff
@@ -12,85 +12,97 @@ trait ConnectionSpec extends BaseSpec {
...
         },
         test("don't track keys") {
-          for {
-            _            <- clientTrackingOff
-            _            <- clientTrackingOn(trackingMode = Some(ClientTrackingMode.OptOut))
-            _            <- clientCaching(false)
-            trackingInfo <- clientTrackingInfo
-          } yield assert(trackingInfo.flags.caching)(isSome(isFalse))
+          ZIO.serviceWithZIO[Redis] { redis =>
+            import redis._
```
mijicd, line 27: Same here.

example/src/main/scala/example/contributorsCache.scala
```diff
@@ -36,12 +36,15 @@ final case class ContributorsCacheLive(r: Redis, s: Sttp) extends ContributorsCa
...
       .ignore
 
   private def read(repository: Repository): ZIO[Redis, ApiError, Contributors] =
-    get(repository.key)
-      .returning[String]
+    ZIO
+      .serviceWithZIO[Redis](
+        _.get(repository.key)
+          .returning[String]
+      )
```
mijicd, line 47:
```suggestion
    ZIO.serviceWithZIO[Redis](_.get(repository.key).returning[String])
```

mijicd (changes requested):
Almost there :)

redis/src/main/scala/zio/redis/RedisCommand.scala
```diff
@@ -30,6 +30,12 @@ final class RedisCommand[-In, +Out] private (val name: String, val input: Input[
       }
       .refineToOrDie[RedisError]
 
+  private[redis] def run(in: In)(implicit executor: RedisExecutor, codec: BinaryCodec): ZIO[Any, RedisError, Out] =
+    executor
+      .execute(resp(in, codec))
+      .flatMap[Any, Throwable, Out](out => ZIO.attempt(output.unsafeDecode(out)(codec)))
+      .refineToOrDie[RedisError]
+
```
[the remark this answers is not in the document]
m-kalai: I don't insist on implicits, it was just a way to get things done faster. I can make them "explicit" args or constructor params. Makes probably more sense as a constructor param as you suggested. Any particular order? Second argument list?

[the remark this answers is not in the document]
mijicd: Sure. It probably makes sense grouping it that, but I'm also fine if you place them in the same one.

redis/src/main/scala/zio/redis/codecs/CRC16.scala
m-kalai: yeah you're right, sorry about that. i renamed it because at certain point i had a collision of codec package and codec method on redis package object, and i forgot about it. i'll keep it then.

redis/src/main/scala/zio/redis/CommandExecutor.scala
m-kalai: yeah i was struggling with naming this one, considering those two methods were making Redis trait before this change. could it be something like `RedisEnvironment` or `CommandExecutable`. Any suggestion?

mijicd: Environment sounds good. They don't need to be implicit anymore either. The question remains: can we remove it completely?

m-kalai: i don't think we can, unless we want to split command creation and execution. If we changed API traits to return `RedisCommand` and pushed invocation of `run` then we wouldn't need these dependencies on those traits. Or do you have some ideas?

redis/src/main/scala/zio/redis/RedisEnvironment.scala
```diff
@@ -0,0 +1,8 @@
+package zio.redis
+
+import zio.schema.codec.BinaryCodec
+
+trait RedisEnvironment {
```
mijicd, line 5: Maybe we should make it package-private. After all, it is an implementation detail.

redis/src/main/scala/zio/redis/RedisCommand.scala
```diff
@@ -20,27 +20,25 @@ import zio._
...
+  private[redis] def resp(in: In): Chunk[RespValue.BulkString] =
     Varargs(StringInput).encode(name.split(" "))(codec) ++ input.encode(in)(codec)
 }
 
 object RedisCommand {
-  private[redis] def apply[In, Out](name: String, input: Input[In], output: Output[Out]): RedisCommand[In, Out] =
-    new RedisCommand(name, input, output)
+  private[redis] def apply[In, Out](name: String, input: Input[In], output: Output[Out])(
+    executor: RedisExecutor,
+    codec: BinaryCodec
```
mijicd, line 41: I'm not sure about this parameter-group. Let's merge them in the first one (sorry :pray: )

mijicd (commented):
Looks really good to me. Two minor things and it's ready to go.

redis/src/main/scala/zio/redis/CommandExecutor.scala
mijicd: I would not "expose" run. As mentioned below, 2nd parameter group is a bit too much as well, imo. On the other side, it makes sense if we restore implicits. :thinking:

redis/src/main/scala/zio/redis/RedisEnvironment.scala
m-kalai: yes, makes sense

redis/src/main/scala/zio/redis/RedisCommand.scala
m-kalai: done

redis/src/main/scala/zio/redis/CommandExecutor.scala
mijicd: That's not bad. We might be able to combine it with `returning` somehow and make a nice "proxy", but let's leave it for the follow-up PR (if you're willing to dig into it :) ).

mijicd:
@m-kalai Please adjust the docs examples and it's ready to go.

mijicd (approved):

mijicd (approved):

PR #738, opened 2023-01-28
[WIP] separation of Redis command and execution

m-kalai:
* if we split command definition and actual execution we're able to remove` executor` and `codec` dependency from `RedisCommand` itself and push it down to actual execution


mijicd (commented):
This PR introduces another, possibly unnecessary abstraction and I have the feeling that it suffers from the same issue as #739: https://github.com/zio/zio-redis/pull/739#pullrequestreview-1274656715

PR #739, opened 2023-01-29
[WIP] exposing RedisEnvironment via environment

m-kalai:
* another approach to pushing `codec` and `executor` dependency from `RedisCommand` creation to it's execution

### DIFF
--- a/redis/src/main/scala/zio/redis/Redis.scala
+++ b/redis/src/main/scala/zio/redis/Redis.scala
@@ -17,7 +17,6 @@
 package zio.redis
 
 import zio._
-import zio.schema.codec.BinaryCodec
 
 trait Redis
     extends api.Connection
@@ -31,14 +30,11 @@ trait Redis
     with api.SortedSets
     with api.Streams
     with api.Scripting
-    with api.Cluster {
-  def codec: BinaryCodec
-  def executor: RedisExecutor
-}
+    with api.Cluster
 
-final case class RedisLive(codec: BinaryCodec, executor: RedisExecutor) extends Redis
+object Redis extends Redis
 
 object RedisLive {
-  lazy val layer: URLayer[RedisExecutor with BinaryCodec, Redis] =
-    ZLayer.fromFunction(RedisLive.apply _)
+  lazy val layer: ULayer[Redis] =
+    ZLayer.succeed(Redis)
 }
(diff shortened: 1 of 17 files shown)


redis/src/main/scala/zio/redis/Redis.scala
```diff
@@ -31,14 +30,11 @@ trait Redis
...
+    with api.Cluster
 
-final case class RedisLive(codec: BinaryCodec, executor: RedisExecutor) extends Redis
+object Redis extends Redis
 
 object RedisLive {
-  lazy val layer: URLayer[RedisExecutor with BinaryCodec, Redis] =
-    ZLayer.fromFunction(RedisLive.apply _)
+  lazy val layer: ULayer[Redis] =
+    ZLayer.succeed(Redis)
```
mijicd, line 39: Well, you may as well remove the `RedisLive` object and pack everything in `Redis`, e.g.:

```scala
object Redis {
  lazy val layer: ULayer[Redis] = ZLayer.succeed(new Redis)
}
```

In the future we may consider renaming it to `live`, perhaps it'll be more consistent with the ecosystem.

redis/src/main/scala/zio/redis/RedisEnvironment.scala
```diff
@@ -1,8 +1,15 @@
...
-  def executor: RedisExecutor
+final case class RedisEnvironment(codec: BinaryCodec, executor: RedisExecutor)
+
+private[redis] object RedisEnvironment {
+  lazy val layer = ZLayer {
+    for {
+      codec    <- ZIO.service[BinaryCodec]
+      executor <- ZIO.service[RedisExecutor]
+    } yield new RedisEnvironment(codec, executor)
+  }
```
mijicd, line 14: Super nitpick:

```suggestion
  lazy val layer = 
    ZLayer {
      for {
        codec    <- ZIO.service[BinaryCodec]
        executor <- ZIO.service[RedisExecutor]
      } yield new RedisEnvironment(codec, executor)
    }
```

mijicd (commented):
A few minor comments from my side, but we have a bigger problem, which is that `RedisEnvironment` (or `BinaryCodec with RedisExecutor`) leaks in user-land, thus going against the service pattern and the changes you did in #727.

Issue #673, opened 2022-11-12
Support Cross-slot command execution


anatolysergeev:
We can add support for some commands like it's been done in [lettuce](https://github.com/lettuce-io/lettuce-core/wiki/Redis-Cluster#cross-slot-command-execution-and-cluster-wide-execution-for-selected-commands)

Regular Redis Cluster commands are limited to single-slot keys operation – either single key commands or multi-key commands that share the same hash slot.

The cross slot limitation can be mitigated by using the advanced cluster API for a set of selected multi-key commands. Commands that operate on keys with different slots are decomposed into multiple commands. The single commands are fired in a fork/join fashion. The commands are issued concurrently to avoid synchronous chaining. Results are synchronized before the command is completed.

Following commands can be supported for cross-slot command execution:

- [ ] DEL: Delete the KEYs. Returns the number of keys that were removed.
- [ ] EXISTS: Count the number of KEYs that exist across the master nodes being responsible for the particular key.
- [ ] MGET: Get the values of all given KEYs. Returns the values in the order of the keys.
- [ ] MSET: Set multiple key/value pairs for all given KEYs. Returns always OK.
- [ ] TOUCH: Alters the last access time of all given KEYs. Returns the number of keys that were touched.
- [ ] UNLINK: Delete the KEYs and reclaiming memory in a different thread. Returns the number of keys that were removed.

Following commands can be executed on multiple cluster nodes operations:

- [ ] CLIENT SETNAME: Set the client name on all known cluster node connections. Returns always OK.
- [ ] KEYS: Return/Stream all keys that are stored on all masters.
- [ ] DBSIZE: Return the number of keys that are stored on all masters.
- [ ] FLUSHALL: Flush all data on the cluster masters. Returns always OK.
- [ ] FLUSHDB: Flush all data on the cluster masters. Returns always OK.
- [ ] RANDOMKEY: Return a random key from a random master.
- [ ] SCAN: Scan the keyspace across the whole cluster according to ReadFrom settings.
- [ ] SCRIPT FLUSH: Remove all the scripts from the script cache on all cluster nodes.
- [ ] SCRIPT LOAD: Load the script into the Lua script cache on all nodes.
- [ ] SCRIPT KILL: Kill the script currently in execution on all cluster nodes. This call does not fail even if no scripts are running.
- [ ] SHUTDOWN: Synchronously save the dataset to disk and then shut down all nodes of the cluster.

PR #754, opened 2023-02-09
add default slot and keys recognition

anatolysergeev:
add default connection for commands without a key and for multiple key commands with the limitation that all keys have to be in the same slot

#673

### DIFF
--- a/redis/src/main/scala/zio/redis/ClusterExecutor.scala
+++ b/redis/src/main/scala/zio/redis/ClusterExecutor.scala
@@ -58,8 +58,8 @@ final case class ClusterExecutor(
     }
 
     for {
-      key    <- ZIO.attempt(command.args(1).value).orElseFail(CusterKeyError)
-      keySlot = Slot((key.asCRC16 % SlotsAmount).toLong)
+      keyOpt <- ZIO.succeed(command.args.collectFirst { case key: RespArgument.Key => key })
+      keySlot = keyOpt.fold(Slot.Default)(key => Slot((key.asCRC16 & (SlotsAmount - 1)).toLong))
       result <- executeSafe(keySlot)
     } yield result
   }
@@ -165,12 +165,8 @@ object ClusterExecutor {
       for (i <- p.slotRange.start to p.slotRange.end) yield Slot(i) -> p.master.address
     }.toMap
 
-  private final val CusterKeyError =
-    RedisError.ProtocolError("Key doesn't found. No way to dispatch this command to Redis Cluster")
   private final val CusterKeyExecutorError =
-    RedisError.IOError(
-      new IOException("Executor for key doesn't found. No way to dispatch this command to Redis Cluster")
-    )
+    RedisError.IOError(new IOException("Executor doesn't found. No way to dispatch this command to Redis Cluster"))
   private final val CusterConnectionError =
     RedisError.IOError(new IOException("The connection to cluster has been failed. Can't reach a single startup node."))
 }
--- a/redis/src/main/scala/zio/redis/RespValue.scala
+++ b/redis/src/main/scala/zio/redis/RespValue.scala
@@ -17,7 +17,6 @@
 package zio.redis
 
 import zio._
-import zio.redis.codecs.CRC16
 import zio.redis.options.Cluster.Slot
 import zio.stream._
 
@@ -75,12 +74,6 @@ object RespValue {
     private[redis] def asString: String = decode(value)
 
     private[redis] def asLong: Long = internal.unsafeReadLong(asString, 0)
-
-    private[redis] def asCRC16: Int = {
-      val betweenBraces = value.dropWhile(b => b != '{').drop(1).takeWhile(b => b != '}')
-      val key           = if (betweenBraces.isEmpty) value else betweenBraces
-      CRC16.get(key)
-    }
   }
 
   final case class Array(values: Chunk[RespValue]) extends RespValue
--- a/redis/src/test/scala/zio/redis/RespArgumentSpec.scala
+++ b/redis/src/test/scala/zio/redis/RespArgumentSpec.scala
@@ -1,0 +1,23 @@
+package zio.redis
+
+import zio.test._
+
+object RespArgumentSpec extends BaseSpec {
+  def spec: Spec[Any, RedisError.ProtocolError] =
+    suite("RespArgument")(
+      suite("BulkString.asCRC16")(
+        test("key without braces") {
+          val key = RespArgument.Key("hello world")
+          assertTrue(15332 == key.asCRC16)
+        },
+        test("key between braces") {
+          val key = RespArgument.Key("hello{key1}wor}ld")
+          assertTrue(41957 == key.asCRC16)
+        },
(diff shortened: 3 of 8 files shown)


redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -526,114 +573,114 @@ object Input {
...
   }
 
   final case class Varargs[-A](input: Input[A]) extends Input[Iterable[A]] {
-    def encode(data: Iterable[A])(implicit codec: BinaryCodec): Chunk[RespValue.BulkString] =
-      data.foldLeft(Chunk.empty: Chunk[RespValue.BulkString])((acc, a) => acc ++ input.encode(a))
+    def encode(data: Iterable[A])(implicit codec: BinaryCodec): RespCommand =
+      data.foldLeft(RespCommand.empty)((acc, a) => acc ++ input.encode(a))
   }
 
   final case class EvalInput[-K, -V](inputK: Input[K], inputV: Input[V]) extends Input[(String, Chunk[K], Chunk[V])] {
```
anatolysergeev, line 632: Controversial idea. What Is interface ideally wanted: inputK would have type KeyInput[K] that inherit Input[K]  and we can translate there any dependency, but we cannot to do that cause we have to have contramap that in the inherited interface that return main trait Input. 🤔 
@mijicd What do you think?

redis/src/main/scala/zio/redis/ClusterExecutor.scala
```diff
@@ -58,8 +58,8 @@ final case class ClusterExecutor(
     }
 
     for {
-      key    <- ZIO.attempt(command(1)).orElseFail(CusterKeyError)
-      keySlot = Slot((key.asCRC16 % SlotsAmount).toLong)
+      keyOpt <- ZIO.succeed(command.respArgs.collectFirst { case key: RespArgument.Key => key })
+      keySlot = keyOpt.fold(Slot.Default)(key => Slot((key.asCRC16 % SlotsAmount).toLong))
```
mijicd, line 62: Since `SlotsAmount` is 2^14, you can replace `%` with `&`, i.e. `key.asCRC16 & (SlotsAmount - 1)`. The same applies for other usages of `%` with `SlotsAmount` (if any).

redis/src/main/scala/zio/redis/RespValue.scala
```diff
@@ -17,12 +17,78 @@
...
+  def apply(respArgs: Chunk[RespArgument]): RespCommand = new RespCommand(respArgs)
+
+  def apply(respArgs: RespArgument*): RespCommand = new RespCommand(Chunk.fromIterable(respArgs))
+
+  def apply(respArg: RespArgument): RespCommand = new RespCommand(Chunk.single(respArg))
+}
+
+sealed trait RespArgument {
+  def respValue: RespValue.BulkString
+}
```
mijicd, line 48: If you would like to proceed with this, please:
- submit it in a separate pull request as it is not related to slots and keys
- either split it to multiple files or respect scala's file-naming convention
- benchmark, as it will introduce some penalty

redis/src/main/scala/zio/redis/RespValue.scala
```diff
@@ -17,12 +17,78 @@
...
+
+object RespArgument {
+
+  case class Unknown(bytes: Chunk[Byte]) extends RespArgument {
```
mijicd, line 52: Case classes should be final.

mijicd (changes requested):
Please split the pull request as suggested, and I'll re-review :pray:

redis/src/main/scala/zio/redis/RespValue.scala
anatolysergeev: I did few benchmarks:
`[info] Benchmark                    (count)   Mode  Cnt   Score   Error  Units
[info] AppendBenchmarks.zio             500  thrpt   30   9.986 ± 0.026  ops/s ## my branch
[info] AppendBenchmarks.zio             500  thrpt   30  10.071 ± 0.035  ops/s

[info] Benchmark          (count)   Mode  Cnt   Score   Error  Units
[info] SetBenchmarks.zio      500  thrpt   30   9.918 ± 0.025  ops/s           ## my branch
[info] SetBenchmarks.zio      500  thrpt   30  10.426 ± 0.037  ops/s

[info] Benchmark          (count)   Mode  Cnt  Score   Error  Units
[info] GetBenchmarks.zio      500  thrpt   30  9.358 ± 0.631  ops/s            ## my branch
[info] GetBenchmarks.zio      500  thrpt   30  9.927 ± 0.023  ops/s

[info] Benchmark           (count)   Mode  Cnt   Score   Error  Units
[info] IncrBenchmarks.zio      500  thrpt   30   9.931 ± 0.203  ops/s          ## my branch
[info] IncrBenchmarks.zio      500  thrpt   30  10.188 ± 0.203  ops/s`

Is it enough?

anatolysergeev: >submit it in a separate pull request as it is not related to slots and keys

It's related to slot and keys, we have to differentiate parameter types in executor

anatolysergeev: >either split it to multiple files or respect scala's file-naming convention

Splitted

mijicd: > >submit it in a separate pull request as it is not related to slots and keys
> 
> It's related to slot and keys, we have to differentiate parameter types in executor

It causes a ripple effect of changes throughout the entire library. I would argue that this is not a "local" change :)

anatolysergeev: I agree with you it is not a "local" change - there is no argues.) I don't know how to solve that issue do not change inner architecture 😇 
But current architecture do not have tools to differentiate Redis (resp) parameter types, that is necessary for advanced client library. If we want to have flexible, easy maintainable executors we need it. Something similar is implemented in the lettuce, i get basic idea from it.

mijicd: I am with you when it comes to the approach to solve it, just do the RespCommand part in one PR and "apply" it another :)

anatolysergeev: Do you want that?) 
https://github.com/zio/zio-redis/pull/763

redis/src/test/scala/zio/redis/RespArgumentSpec.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import zio.test._
+
+object RespArgumentSpec extends BaseSpec {
+  def spec: Spec[Any, RedisError.ProtocolError] =
+    suite("RespArgument")(
+      suite("BulkString.asCRC16")(
+        test("key without braces") {
+          val key = RespArgument.Key("hello world")
+          assertTrue(15332 == key.asCRC16)
```
mijicd, line 11: Nitpick (I'm fine with leaving it as it is as well):

```suggestion
          assertTrue(15332 == RespArgument.Key("hello world").asCRC16)
```

mijicd (approved):

PR #763, opened 2023-02-25
add resp command in order to distinguish resp types

anatolysergeev:


redis/src/main/scala/zio/redis/RespArgument.scala
```diff
@@ -0,0 +1,64 @@
...
+
+object RespArgument {
+
+  final case class Unknown(bytes: Chunk[Byte]) extends RespArgument {
+    lazy val respValue: BulkString = RespValue.BulkString(bytes)
+  }
+
+  object Unknown {
+    def apply(str: String): Unknown                                            = Unknown(Chunk.fromArray(str.getBytes(StandardCharsets.UTF_8)))
+    def apply[A](data: A)(implicit codec: BinaryCodec, schema: Schema[A]): Key = Key(codec.encode(schema)(data))
```
mijicd, line 38: This doesn't belong here.

redis/src/main/scala/zio/redis/RespCommand.scala
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package zio.redis
+
+import zio.Chunk
+
+final case class RespCommand(respArgs: Chunk[RespArgument]) {
+  def ++(that: RespCommand): RespCommand = RespCommand(this.respArgs ++ that.respArgs)
+
+  def transformArgs(f: RespArgument => RespArgument): RespCommand = RespCommand(respArgs.map(f(_)))
```
mijicd, line 24: Let's rename to `mapArguments`.

redis/src/main/scala/zio/redis/RespCommand.scala
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.redis
+
+import zio.Chunk
+
+final case class RespCommand(respArgs: Chunk[RespArgument]) {
```
mijicd, line 21: I'd rename it to `args` or `arguments`. No need for carrying the prefix.

redis/src/main/scala/zio/redis/RespCommand.scala
```diff
@@ -0,0 +1,36 @@
...
+
+object RespCommand {
+
+  def empty: RespCommand = new RespCommand(Chunk.empty)
+
+  def apply(respArgs: Chunk[RespArgument]): RespCommand = new RespCommand(respArgs)
+
+  def apply(respArgs: RespArgument*): RespCommand = new RespCommand(Chunk.fromIterable(respArgs))
+
+  def apply(respArg: RespArgument): RespCommand = new RespCommand(Chunk.single(respArg))
```
mijicd, line 35: Let's drop the prefix here as well.

redis/src/main/scala/zio/redis/RespArgument.scala
```diff
@@ -0,0 +1,64 @@
...
+
+import zio.Chunk
+import zio.redis.RespValue.BulkString
+import zio.schema.Schema
+import zio.schema.codec.BinaryCodec
+
+import java.nio.charset.StandardCharsets
+
+sealed trait RespArgument {
+  def respValue: RespValue.BulkString
```
mijicd, line 27: Let's drop the prefix.

redis/src/main/scala/zio/redis/RespArgument.scala
```diff
@@ -0,0 +1,64 @@
...
+    def apply[A](data: A)(implicit codec: BinaryCodec, schema: Schema[A]): Key = Key(codec.encode(schema)(data))
+  }
+
+  final case class CommandName(str: String) extends RespArgument {
+    lazy val respValue: BulkString = RespValue.bulkString(str)
+  }
+
+  final case class Literal(str: String) extends RespArgument {
+    lazy val respValue: BulkString = RespValue.bulkString(str)
+  }
```
mijicd, line 47: Just to clarify, `Value` is the generic value holder (i.e., anything that can be encoded as chunk of bytes) while `Literal` is to be used mostly internally wherever we're dealing with string literals (e.g. command options) ?

redis/src/main/scala/zio/redis/Input.scala
```diff
@@ -27,428 +26,477 @@ import java.util.concurrent.TimeUnit
...
-    def encode(data: A)(implicit codec: BinaryCodec): Chunk[RespValue.BulkString] =
-      Chunk.single(encodeBytes(data))
+    def encode(data: A)(implicit codec: BinaryCodec): RespCommand =
+      RespCommand(RespArgument.Value(data))
   }
 
-  case object ByteInput extends Input[Chunk[Byte]] {
-    def encode(data: Chunk[Byte])(implicit codec: BinaryCodec): Chunk[RespValue.BulkString] =
-      Chunk.single(RespValue.BulkString(data))
+  final case class ArbitraryKeyInput[A: Schema]() extends Input[A] {
```
mijicd, line 342: Shall we rename the `ArbitraryInput` to `ArbitraryValueInput` for symmetry?

mijicd (commented):
Just a few remarks from me, otherwise, looks neat!

redis/src/main/scala/zio/redis/RespArgument.scala
anatolysergeev: Yeah, But may be I thought about `Value` more like something that we want to save in redis and that's why there are we have `Unknown` case - where i don't know is it have to have some separate case for example like group-name or consumer or it should be a `Value` like ip address or ttl duration 🤔

redis/src/main/scala/zio/redis/RespArgument.scala
anatolysergeev: Ups, it should return Unknown 😇

mijicd (approved):

PR #773, opened 2023-03-16
ADD zio-redis-embedded module

ioleo:
This PR introduces `zio-redis-embedded` module for integration testing with embedded redis instance.

See README for details.


mijicd (approved):
It might be an interesting approach to verify it against our default test suite, but I guess that'd require some SBT module "juggling".

mijicd (approved):

PR #801, opened 2023-04-04
Support 'GET' option in 'SET' command

### LINKED ISSUE #222: Add GET option to the SET command
`SET` command got `GET` optional parameter in Redis version 6.2. This option will also affect the format of the command output. For more information go [here](https://redis.io/commands/set).

**Note:** This one is not the priority. Let's wait for it to be stable enough and then we can add it to `zio-redis`.

drmarjanovic:
Closes #222

### DIFF
--- a/modules/redis/src/test/scala/zio/redis/KeysSpec.scala
+++ b/modules/redis/src/test/scala/zio/redis/KeysSpec.scala
@@ -26,15 +26,7 @@ trait KeysSpec extends BaseSpec {
           v     <- redis.setGet(key, value)
         } yield assert(v)(isNone)
       },
-      test("setGet with the existing key and int value") {
-        for {
-          redis <- ZIO.service[Redis]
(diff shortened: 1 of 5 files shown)


modules/redis/src/test/scala/zio/redis/KeysSpec.scala
```diff
@@ -18,6 +18,32 @@ trait KeysSpec extends BaseSpec {
...
+      test("setGet with the existing key and string value") {
+        for {
+          redis    <- ZIO.service[Redis]
+          value    <- uuid
+          key      <- uuid
+          _        <- redis.set(key, value)
+          newValue <- uuid
+          v        <- redis.setGet(key, newValue)
+        } yield assert(v)(isSome(equalTo(value)))
+      },
```
mijicd, line 46: These are redundant.

mijicd (approved):

PR #815, opened 2023-04-10
Unify 'Value' and 'Unknown' command arguments

drmarjanovic:
### Description: 

- Instead of `Unknown` and `Value` use `Value` only. 
- Sorted `Input` alphabetically. 
- Fixed minor typos.


mijicd:
@anatolysergeev Please take a look at this PR, I'm not sure whether or not it matches your idea about command arguments.

anatolysergeev:
I add Unknown value because i don't have, and check all commands, and don't know how that can affect, everything, that's why i was honest with my self and types to do in Unknown. 
If we want to delete that, It's ok with me.

anatolysergeev (approved):

Issue #804, opened 2023-04-06
Upgrade to Redis 7


mijicd:
Adjust the current API to accommodate the changes version 7 introduced. Note the missing functionality and update the existing tickets (or create new ones) accordingly.

mijicd:
cc @drmarjanovic @mvelimir

PR #824, opened 2023-04-15
Add 'WithScore' option both to 'zRank' and 'zRevRank' commands

drmarjanovic:
### Description:

**Part of #804**

- Updated `docker-compose` to use Redis 7.2.
- Provide `WithScore` flag.
- Provided `zRankWithScore` and `zRevRankWithScore`. 
- Adapted documentation.
- Used `WithScoresInput` instead of `ArbitraryValueInput[String]()`. 
- Replaced `STRALGO LCS` with `LCS`. Removed support for strings. Changed API.  Reason: _"Remove STRALGO command, preserve LCS a standalone command which only works on keys"_.  More info [here](https://raw.githubusercontent.com/redis/redis/7.0/00-RELEASENOTES).

### DIFF
--- a/modules/redis/src/main/scala/zio/redis/options/SortedSets.scala
+++ b/modules/redis/src/main/scala/zio/redis/options/SortedSets.scala
@@ -82,6 +82,8 @@ trait SortedSets {
 
   type MemberScores[+M] = Chunk[MemberScore[M]]
 
+  sealed case class RankScore(score: Double, rank: Long)
+
   sealed trait ScoreMaximum { self =>
     private[redis] final def asString: String =
       self match {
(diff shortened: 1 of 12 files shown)


modules/redis/src/main/scala/zio/redis/options/SortedSets.scala
```diff
@@ -82,6 +82,8 @@ trait SortedSets {
 
   type MemberScores[+M] = Chunk[MemberScore[M]]
 
+  sealed case class RankScore(score: Double, rank: Long)
```
mijicd, line 85:
```suggestion
  sealed case class RankScore(rank: Long, score: Double)
```

We should consider introducing newtypes (maybe through zio-prelude) to represent things like `Rank` consistently and "cost-free". That's a topic for a separate PR, though.

mijicd (approved):
LGTM with minor remark

mijicd (approved):

PR #835, opened 2023-05-29
Handle empty bulk string (#808)

### LINKED ISSUE #808: Handle empty content properly
Reading a key containing an empty string returns `None` instead of the expected `Some("")`.

sunLightLeaf:
Resolves #808 

This did change how `getRange` works, as it now returns empty quotes even when the sought range exceed the length of the string, but to my knowledge (tweaking around with redis-cli) `getRange` returning empty quotes in such scenarios is standard redis behaviour. 

Hope that helps! I'm ready to tinker with this some more in case you think this solution could be improved.

### DIFF
--- a/modules/redis/src/test/scala/zio/redis/StringsSpec.scala
+++ b/modules/redis/src/test/scala/zio/redis/StringsSpec.scala
@@ -989,7 +989,7 @@ trait StringsSpec extends BaseSpec {
             result <- redis.get(key).returning[String]
           } yield assert(result)(isNone)
         },
-        test("emtpy string") {
+        test("empty string") {
           for {
             redis  <- ZIO.service[Redis]
             key    <- uuid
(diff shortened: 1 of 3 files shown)


mijicd:
@sunLightLeaf Please fix formatting errors by running `sbt prepare`.

sunLightLeaf:
Strange. Before running `sbt prepare` the only failed check [was](https://github.com/zio/zio-redis/actions/runs/5107240028) lint, but now, after fixing the formatting, four other checks have turned red. Formatting alone shouldn't have had such an impact.

mijicd:
@sunLightLeaf it isn't related to formatting. My guess is that they're just flaky.

modules/redis/src/test/scala/zio/redis/StringsSpec.scala
```diff
@@ -982,13 +982,21 @@ trait StringsSpec extends BaseSpec {
...
         },
-        test("emtpy string") {
+        test("non existent string") {
           for {
             redis  <- ZIO.service[Redis]
             key    <- uuid
             result <- redis.get(key).returning[String]
           } yield assert(result)(isNone)
         },
+        test("emtpy string") {
```
mijicd, line 992:
```suggestion
        test("empty string") {
```

mijicd (approved):

mijicd (approved):

PR #852, opened 2023-06-27
Explicit resp command name

0pg:
Split `RespCommand` into `RespCommandName` and `RespCommandArguments`

### DIFF
--- a/modules/redis/src/main/scala/zio/redis/internal/RespCommand.scala
+++ b/modules/redis/src/main/scala/zio/redis/internal/RespCommand.scala
@@ -18,18 +18,32 @@ package zio.redis.internal
 
 import zio.Chunk
 
-private[redis] final case class RespCommand(args: Chunk[RespCommandArgument]) extends AnyVal {
-  def ++(that: RespCommand): RespCommand = RespCommand(this.args ++ that.args)
-
-  def mapArguments(f: RespCommandArgument => RespCommandArgument): RespCommand = RespCommand(args.map(f(_)))
+private[redis] final case class RespCommand(name: RespCommandName, args: RespCommandArguments) {
+  def bulkStrings: Chunk[RespValue.BulkString] = name.bulkStrings ++ args.bulkStrings
 }
 
-private[redis] object RespCommand {
-  def empty: RespCommand = new RespCommand(Chunk.empty)
+private[redis] final case class RespCommandName(str: String) extends AnyVal {
+  def bulkStrings: Chunk[RespValue.BulkString] = Chunk.fromArray(str.split(" ").map(RespValue.bulkString))
+}
 
-  def apply(args: Chunk[RespCommandArgument]): RespCommand = new RespCommand(args)
+private[redis] final case class RespCommandArguments(values: Chunk[RespCommandArgument]) extends AnyVal {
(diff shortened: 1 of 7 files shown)


0pg:
@anovakovic01 about discussion in https://github.com/zio/zio-redis/pull/714#discussion_r1241109041
Please feel free to give any ideas when you are free 😃

modules/redis/src/main/scala/zio/redis/internal/RespCommand.scala
```diff
@@ -18,18 +18,32 @@ package zio.redis.internal
...
+}
+
+private[redis] final case class RespCommandName(str: String) extends AnyVal {
+  def bulkStrings: Chunk[RespValue.BulkString] = Chunk.fromArray(str.split(" ").map(RespValue.bulkString))
+}
+
+private[redis] final case class RespCommandArguments(values: Chunk[RespCommandArgument]) extends AnyVal {
+  def ++(that: RespCommandArguments): RespCommandArguments = RespCommandArguments(this.values ++ that.values)
+
+  def mapArguments(f: RespCommandArgument => RespCommandArgument): RespCommandArguments = RespCommandArguments(
```
anovakovic01, line 32: This method can be named `map` (IMO it's clear what you're mapping).

modules/redis/src/main/scala/zio/redis/internal/RespCommand.scala
```diff
@@ -18,18 +18,32 @@ package zio.redis.internal
...
+    values.map(f(_))
+  )
+
+  def buildCommand(name: RespCommandName): RespCommand = RespCommand(name, this)
```
anovakovic01, line 36: Do you need this method? It looks like it's reimplementation of the `RespCommand` constructor which you already have.

modules/redis/src/main/scala/zio/redis/internal/RespCommand.scala
```diff
@@ -18,18 +18,32 @@ package zio.redis.internal
...
+private[redis] final case class RespCommand(name: RespCommandName, args: RespCommandArguments) {
+  def bulkStrings: Chunk[RespValue.BulkString] = name.bulkStrings ++ args.bulkStrings
+}
+
+private[redis] final case class RespCommandName(str: String) extends AnyVal {
+  def bulkStrings: Chunk[RespValue.BulkString] = Chunk.fromArray(str.split(" ").map(RespValue.bulkString))
+}
+
+private[redis] final case class RespCommandArguments(values: Chunk[RespCommandArgument]) extends AnyVal {
+  def ++(that: RespCommandArguments): RespCommandArguments = RespCommandArguments(this.values ++ that.values)
```
anovakovic01, line 30: Is this method used anywhere now?

anovakovic01 (changes requested):

modules/redis/src/main/scala/zio/redis/internal/RespCommand.scala
0pg: not necessarily, just helper for builder-like interface. I removed and replaced using constructor

modules/redis/src/main/scala/zio/redis/internal/RespCommand.scala
0pg: yes it's used to construct `Input` subtype instances

PR #881, opened 2023-09-07
Update Scala versions, ZIO, sbt and zio-sbt plugin

guizmaii:
Closes #878, closes #879

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,6 +5,10 @@ enablePlugins(ZioSbtEcosystemPlugin, ZioSbtCiPlugin)
 inThisBuild(
   List(
     name              := "ZIO Redis",
+    zioVersion        := "2.0.16",
+    scala212          := "2.12.18",
+    scala213          := "2.13.11",
+    scala3            := "3.3.0",
     ciEnabledBranches := List("master"),
     ciExtraTestSteps  := List(
       SingleStep(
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -20,7 +24,6 @@ inThisBuild(
         run = Some("sbt ++${{ matrix.scala }} IntegrationTest/test")
       )
     ),
-    crossScalaVersions -= scala211.value,
```
guizmaii, line 23: zio-sbt doesn't provide support for Scala 2.11 anymore so we don't need this configuration anymore

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version = 1.8.3
+sbt.version = 1.9.4
```
mijicd, line 1: Will it cause deprecation warnings (errors) due to the usage of integration test config?

guizmaii: Yes, it does trigger warnings

build.sbt
```diff
@@ -5,6 +5,10 @@ enablePlugins(ZioSbtEcosystemPlugin, ZioSbtCiPlugin)
 inThisBuild(
   List(
     name              := "ZIO Redis",
+    zioVersion        := "2.0.16",
+    scala212          := "2.12.18",
+    scala213          := "2.13.11",
+    scala3            := "3.3.0",
```
mijicd, line 11: These are yet to be upgraded in the plugin, right?

mijicd (approved):

build.sbt
guizmaii: Not sure. I think I updated them recently in zio-sbt but I think you shouldn't have to wait for zio-sbt to update the Scala versions and make a new release for you to update yours

PR #902, opened 2023-10-20
add support for the COPY command

mberndt123:
Hi,

it seems like the [COPY](https://redis.io/commands/copy/) command is not currently supported, so I've added it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
     continue-on-error: true
     steps:
     - name: Git Checkout
-      uses: actions/checkout@v4.1.1
+      uses: actions/checkout@v4.1.0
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     continue-on-error: true
     steps:
     - name: Git Checkout
-      uses: actions/checkout@v4.1.1
```
mijicd, line 23: Actions are auto-generated and updated through zio-sbt, so I'd prefer not to edit them manually.

mijicd (dismissed):
Thanks for taking care of this! Just one thing to clarify, otherwise it's good to go.

mijicd (approved):

PR #980, opened 2024-07-28
Rename zio-redis-it's BaseSpec to avoid confusion with test's BaseSpec

maizy:
Would it be correct to rename `zio.redis.BaseSpec` inside `modules/redis-it/src/test/scala/zio/redis/BaseSpec.scala`?

If so, this PR does it.

The problem is that IntelliJ IDEA imports sbt projects incorrectly and adds an implicit `test->test` dependency to all dependent modules. This leads to confusing `zio.redis.BaseSpec` in `modules/redis/src/test/scala/zio/redis/BaseSpec.scala` and `modules/redis-it/src/test/scala/zio/redis/BaseSpec.scala`. Also, test's BaseSpec wins over it's BaseSpec.

This problem only affects IJ IDEA, for example VSCode + Metals works as expected.

An alternative solution might be to rename `zio.redis` package for itegration tests to something like `zio.redis_it`.

### DIFF
--- a/modules/redis-it/src/test/scala/zio/redis/ApiSpec.scala
+++ b/modules/redis-it/src/test/scala/zio/redis/ApiSpec.scala
@@ -24,9 +24,9 @@ object ApiSpec
     suite("Redis commands")(ClusterSuite, SingleNodeSuite)
       .provideShared(
         compose(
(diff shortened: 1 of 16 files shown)


mijicd (commented):
Let's rename to IntegrationSpec then. Otherwise LGTM!

mijicd (approved):

PR #930, opened 2024-01-30
Add support SSL/TLS connection (832)

gkfabs:
Hi,

I added support for [SSL/TLS](https://github.com/zio/zio-redis/issues/832)

PR #984, opened 2024-08-04
fix infinity loop in RedisUri, add tests for ClusterPartitionOutput

### LINKED ISSUE #970: Infinite loop in RedisUri
This line creates an infinite loop which is breaking tests.

https://github.com/zio/zio-redis/blob/941013ab5c95a05beb0269b3ef1d322bb9225875/modules/redis/src/main/scala/zio/redis/RedisUri.scala#L29

maizy:
fixes #970

Bug in RedisUri was introduced when adding TLS support (#930). `RedisUri.apply(String)` constructor called itself after building connection URI string, causing cluster spec to hang.

Unfortunately I haven't managed to run cluster spec locally yet, but now it's another problem, possibly with NATted network environment inside the test cluster.

### DIFF
--- a/modules/redis-it/src/test/scala/zio/redis/ClusterSpec.scala
+++ b/modules/redis-it/src/test/scala/zio/redis/ClusterSpec.scala
@@ -18,7 +18,7 @@ trait ClusterSpec extends IntegrationSpec {
                 .foreach(0 to 5) { n =>
                   ZIO
                     .attempt(docker.getServiceHost(s"cluster-node$n", port))
-                    .map(host => RedisUri(s"$host:$port"))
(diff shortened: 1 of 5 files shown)


maizy:
Looks like the old `docker-compose` doesn't exist in ubuntu-latest anymore.

I'll try to update testcontainers, as far as I know the new version knows how to use `docker compose` plugin.

maizy:
`docker-compose` was removed: https://github.com/actions/runner-images/issues/9692

maizy:
I've made a separate PR to update testcontainers #985

mijicd:
@maizy please rebase

mijicd (approved):

Issue #1000, opened 2024-09-26
Extend Streams API


drmarjanovic:
The following commands are missing in Streams API:
- [x] `XAUTOCLAIM`: The [command](https://redis.io/docs/latest/commands/xautoclaim/) is available since 6.2.0
- [x] `XSETID`: The [command](https://redis.io/docs/latest/commands/xsetid/) is available since 5.0.0

The following commands should be adjusted or extended:
- [x] `XADD`: [docs](https://redis.io/docs/latest/commands/xadd/)
  - [x] 6.2.0: Added `NOMKSTREAM`, `MINID`, and `LIMIT` options
  - [x] 7.0.0: Added support for the `<ms>-*` explicit ID form
- [x] `XCLAIM`: The [command](https://redis.io/docs/latest/commands/xclaim/) is missing the `LASTID` option
- [x] `XGROUP CREATE`: [docs](https://redis.io/docs/latest/commands/xgroup-create/)
  -  Redis 7.0.0: Added `ENTRIESREAD entries-read` option
  -  Currently supported `id` but should be extended with `<id | $>`
- [x] `XGROUP SETID`: [docs](https://redis.io/docs/latest/commands/xgroup-setid/)
  - 7.0.0 Added `ENTRIESREAD entries-read` option
  - Currently supported `id` but should be extended with `<id | $>`
- [x] `XINFO CONSUMERS`: [docs](https://redis.io/docs/latest/commands/xinfo-consumers/)
  - 7.2.0: Added `inactive` field in the output
- [x] `XINFO GROUPS`: [docs](https://redis.io/docs/latest/commands/xinfo-groups/)
  - 7.0.0: Added `entries-read` and `lag` fields in the output
- [x] `XINFO STREAM`: [docs](https://redis.io/docs/latest/commands/xinfo-stream/)
  - 7.0.0: Added `max-deleted-entry-id`, `entries-added`, `recorded-first-entry-id`, `entries-read` and `lag` fields
  - 7.2.0: Added `active-time` field, and changed the meaning of `seen-time`
- [x] `XTRIM`: The [command](https://redis.io/docs/latest/commands/xtrim/) is missing `MAXLEN | MINID` and `threshold` options

mijicd:
@hcwilhelm tackled the 6.2.x part of `XADD` in #1039

mijicd:
[...]

PR #1033, opened 2025-02-14
Extends streams api#1000

hcwilhelm:
PR to Fix Extend Streams API #1000

- XAUTOCLAIM was not missing, so no action is needed here.
- XSETID is, according to the Redis docs, an internal command. So I am not sure if we need it in the streams API.


guizmaii:
@hcwilhelm Could you please sign the CLA? 🙂

hcwilhelm:
> @hcwilhelm Could you please sign the CLA? 🙂

I have accepted the CLA several times now, but I think I know what's wrong. The Email address in my commits is not correct

PR #1039, opened 2025-02-18
Streams api#1000

hcwilhelm:
PR to Fix Extend Streams API #1000

- XAUTOCLAIM was not missing, so no action is needed here.
- XSETID is, according to the Redis docs, an internal command. So I am not sure if we need it in the streams API.

### DIFF
--- a/modules/redis/src/main/scala/zio/redis/api/Streams.scala
+++ b/modules/redis/src/main/scala/zio/redis/api/Streams.scala
@@ -130,24 +130,37 @@ trait Streams[G[+_]] extends RedisEnvironment[G] {
   ): ResultBuilder1[Option, G] =
     new ResultBuilder1[Option, G] {
       def returning[R: Schema]: G[Option[Id[R]]] = {
-        val command = RedisCommand(
-          XAdd,
-          Tuple5(
-            ArbitraryKeyInput[SK](),
-            OptionalInput(NoMkStreamInput),
-            OptionalInput(CappedStreamInput),
-            ArbitraryValueInput[I](),
-            NonEmptyList(Tuple2(ArbitraryKeyInput[K](), ArbitraryValueInput[V]()))
-          ),
-          OptionalOutput(ArbitraryOutput[R]())
-        )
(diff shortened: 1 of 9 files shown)


hcwilhelm:
@guizmaii @mijicd Sorry guys, I completely messed up my other PR when trying to rewrite the commits to have the correct email address. So I needed to close that PR and create this new one. 😓

mijicd (commented):
Few minor remarks, otherwise LGTM

hcwilhelm:
@mijicd Ok, I addressed your comments ;)

modules/redis/src/main/scala/zio/redis/api/Streams.scala
```diff
@@ -64,32 +64,34 @@ trait Streams[G[+_]] extends RedisEnvironment[G] {
...
    *   field and value pair
    * @param pairs
    *   rest of the field and value pairs
    * @return
    *   ID of the added entry.
    */
-  final def xAdd[SK: Schema, I: Schema, K: Schema, V: Schema](
-    key: SK,
-    id: I,
+  final def xAdd[SK: Schema, I: Schema, K: Schema, V: Schema](key: SK, id: I, noMakeStream: Boolean = false)(
```
mijicd, line 76: @hcwilhelm sorry I missed this one, let's please preserve the compatibility here, i.e. no need for two argument groups.

mijicd (commented):
Sorry for another nitpick, I missed it in the first pass.

modules/redis/src/main/scala/zio/redis/api/Streams.scala
hcwilhelm: @mijicd I think this is not possible because the compiler does not accept default parameters in a signature with a *-parameter. 

```
a parameter section with a `*`-parameter is not allowed to have default arguments
  final def xAdd[SK: Schema, I: Schema, K: Schema, V: Schema](
```


or do you have another suggestion on how to solve this?

mijicd (approved):

PR #1059, opened 2025-03-03
Add `PING` support

### LINKED ISSUE #919: Re-Add `PING` connection command
During #825, the `PING` command, alongside other connection commands, was removed since it was considered useless.

At the moment, said command would be useful for my company's health endpoints, as it can attest the healthy communication of each API with Redis.

Since there was no explicit reason for those commands to be removed, would this topic be pertinent? Why was `PING` removed?

Thanks in advance for you time!

guizmaii:
Resolves #919 

See PING doc: https://redis.io/docs/latest/commands/ping/

### DIFF
--- a/modules/redis/src/main/scala/zio/redis/api/Connection.scala
+++ b/modules/redis/src/main/scala/zio/redis/api/Connection.scala
@@ -106,7 +106,7 @@ trait Connection[G[+_]] extends RedisEnvironment[G] {
   /**
    * @param message
    *   expected response payload
-   *  
+   *
    * @return
    *   PONG if no argument is provided, otherwise return a copy of the argument
    */
--- a/modules/redis/src/main/scala/zio/redis/api/Connection.scala
+++ b/modules/redis/src/main/scala/zio/redis/api/Connection.scala
@@ -21,8 +21,6 @@ import zio.redis.Output._
 import zio.redis._
 import zio.redis.internal.{RedisCommand, RedisEnvironment}
 
-import java.nio.charset.StandardCharsets
-
 trait Connection[G[+_]] extends RedisEnvironment[G] {
   import Connection.{Auth => _, _}
 
@@ -113,9 +111,7 @@ trait Connection[G[+_]] extends RedisEnvironment[G] {
   final def ping(message: Option[String] = None): G[String] =
     message match {
       case None        => RedisCommand(Ping, NoInput, StringOutput).run(())
-      case Some(value) =>
-        RedisCommand(Ping, StringInput, BulkStringOutput.map(v => new String(v.toArray, StandardCharsets.UTF_8)))
-          .run(value)
+      case Some(value) => RedisCommand(Ping, StringInput, MultiStringOutput).run(value)
     }
 
   /**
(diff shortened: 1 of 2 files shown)


modules/redis/src/main/scala/zio/redis/api/Connection.scala
```diff
@@ -101,6 +101,15 @@ trait Connection[G[+_]] extends RedisEnvironment[G] {
     command.run(name)
   }
 
+  /**
+   * https://redis.io/docs/latest/commands/ping/
+   */
```
mijicd, line 106:
```suggestion
  /**
   * @param message
   *   expected response payload
   *  
   * @return
   *   PONG if no argument is provided, otherwise return a copy of the argument
   */
```

mijicd (commented):
One mega nit-pick (not that I'm super enthusiastic about scaladoc, just for consistency).

guizmaii:
@mijicd Any idea what makes the second test fail:
```scala
        - Redis commands - Single node executor - connection - ping - with an argument, returns the argument
          Exception in thread "zio-fiber-724" zio.redis.RedisError$ProtocolError: null
          	at zio.redis.internal.RedisEnvironment.RunOps.run(RedisEnvironment.scala:31)
          	at zio.redis.ConnectionSpec.connectionSuite(ConnectionSpec.scala:41)
          	at zio.redis.ConnectionSpec.connectionSuite(ConnectionSpec.scala:39)
```

🤔

mijicd:
> @mijicd Any idea what makes the second test fail:
> 
> ```scala
>         - Redis commands - Single node executor - connection - ping - with an argument, returns the argument
>           Exception in thread "zio-fiber-724" zio.redis.RedisError$ProtocolError: null
>           	at zio.redis.internal.RedisEnvironment.RunOps.run(RedisEnvironment.scala:31)
>           	at zio.redis.ConnectionSpec.connectionSuite(ConnectionSpec.scala:41)
>           	at zio.redis.ConnectionSpec.connectionSuite(ConnectionSpec.scala:39)
> ```
> 
> 🤔

@guizmaii I think you will need to tweak the message case to use the `MultiStringOutput`.

guizmaii:
@mijicd  

> I think you will need to tweak the message case to use the `MultiStringOutput`.

Just tried this but the first test doesn't pass with this code.
image

I'll use the `MultiStringOutput` for the second case so I don't have to `.map(... new String ...)`
Thanks! Didn't see the `MultiStringOutput`

mijicd:
> @mijicd  
> 
> > I think you will need to tweak the message case to use the `MultiStringOutput`.
> 
> Just tried this but the first test doesn't pass with this code.
> image
> 
> I'll use the `MultiStringOutput` for the second case so I don't have to `.map(... new String ...)`
> Thanks! Didn't see the `MultiStringOutput`

Yeah, sorry, I meant for the second case (I wrote "message case"). Befpre we had a dedicated output for this command, but it would be an overhead.

Thanks for taking care of this!

guizmaii:
> Thanks for taking care of this!

Pleasure :)

I need it

mijicd (approved):

PR #1065, opened 2025-03-04
Add missing sortedSet commands and update ZRange command

hcwilhelm:
Implement missing SortedSets commands and update ZRange command.

### DIFF
--- a/modules/redis/src/main/scala/zio/redis/api/SortedSets.scala
+++ b/modules/redis/src/main/scala/zio/redis/api/SortedSets.scala
@@ -28,6 +28,195 @@ trait SortedSets[G[+_]] extends RedisEnvironment[G] {
   import SortedSets._
 
   /**
+   * Pops one or more elements, that are member-score pairs, from the first non-empty sorted set in the provided list of
+   * key names.
+   *
+   * @param count
+   *  The optional COUNT can be used to specify the number of elements to pop
+   * @param key
+   *  Key of the set
+   * @param keys
+   *  Keys of the rest sets
+   * @return
+   *  A two-element array with the first element being the name of the key from which elements were popped, and the
+   *  second element is an array of the popped elements. Every entry in the elements array is also an array that
+   *  contains the member and its score.
+   */
+  final def zmPopMin[K: Schema](
+    count: Option[Long] = None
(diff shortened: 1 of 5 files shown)


hcwilhelm:
@guizmaii @mijicd, please review this should fix the complete sortedSets ticket

modules/redis/src/main/scala/zio/redis/api/SortedSets.scala
```diff
@@ -580,6 +769,164 @@ trait SortedSets[G[+_]] extends RedisEnvironment[G] {
...
+   *   ZRANGE can perform different types of range queries: by index (rank), by the score, or by lexicographical order.
+   * @param rev
+   *   The optional REV argument reverses the ordering, so elements are ordered from highest to lowest score, and
+   *   score ties are resolved by reverse lexicographical ordering.
+   * @param limit
+   *   The optional LIMIT argument can be used to obtain a sub-range from the matching elements (similar to SELECT LIMIT offset, count in SQL)
+   * @return
+   *   A list of members in the specified range with, optionally, their scores when the WITHSCORES option is given.
+   */
+  final def zRange[K: Schema](
```
mijicd, line 787: Since we're adding 7.x API, maybe we should remove the old `zRange` function and all "deprecated" functions (or at least "implement" them in terms of `zRange` and mark as `deprecated`)? WDYT?

mijicd (commented):
One minor remark, otherwise LGTM!

modules/redis/src/main/scala/zio/redis/api/SortedSets.scala
hcwilhelm: @mijicd Ok I deprecated the methods and implemented then in terms of the new zRange command, but CI/CD is failing because of the use of deprecated methods 🤔  How do I fix that ?

mijicd: I think the only way to fix that (and keep the deprecation notes) is to live up to our suggestions, a.k.a. update affected calls to new `zRange`. Alternatively, we could live with ignoring warnings, but that's just postponing the previous approach.

mijicd: Just change the calls in tests, don't delete.

PR #1144, opened 2025-10-11
Add basic reconnect method to RedisConnection

gkfabs:
Maybe I should add a back pressure retry scheduler in `SingleNodeExecutor.onError` configurable with some new variables in `RedisConfig` ?


gkfabs:
Do you think you could release a new version with this change? Thank you very much

PR #1250, opened 2026-04-21
Send AUTH on subscription connections - fixes #1249

### LINKED ISSUE #1249: SingleNodeSubscriptionExecutor never sends AUTH — subscriptions fail against password-protected Redis
## Summary
`SingleNodeSubscriptionExecutor` opens its own TCP connection separate from the
command executor (Redis pub/sub cannot share a socket with normal commands).
That connection never issues `AUTH`, even when `RedisConfig.auth` is configured.
Every subscribe attempt against a Redis with `requirepass`/ACLs fails with
`NOAUTH Authentication required.` and `SingleNodeRunner` reconnects forever.

#1205 (closes #1149) added `AUTH` to `SingleNodeExecutor` and `ClusterExecutor`
but left `SingleNodeSubscriptionExecutor` untouched. The tests added in that
PR only exercise `Redis.singleNode`, not `RedisSubscription.singleNode`.

## Environment
- `zio-redis` 1.2.1 (current latest)
- Verified on Scala 3.3.7 / JVM 21

## Reproducer
Start a password-protected Redis:
```
docker run --rm -p 6379:6379 redis:7 --requirepass devpw
```

Save as `repro.scala` and run with `scala-cli run repro.scala`:
```scala
//> using scala 3.3.7
//> using dep dev.zio::zio-redis:1.2.1
//> using dep dev.zio::zio-schema:1.7.5

import zio.*
import zio.redis.*

object Repro extends ZIOAppDefault:

  private val config = RedisConfig(
    host = "localhost",
    port = 6379,
    auth = Some(RedisConfig.Auth("devpw")),
  )

  private val layer: ZLayer[Any, RedisError.IOError, Redis & RedisSubscription] =
    (ZLayer.succeed[CodecSupplier](CodecSupplier.utf8) ++ ZLayer.succeed(config)) >>>
      (Redis.singleNode ++ RedisSubscription.singleNode)

  def run =
    val program =
      for
        redis        <- ZIO.service[Redis]
        subscription <- ZIO.service[RedisSubscription]
        _            <- redis.publish("ch", "hi").unit
        _            <- ZIO.logInfo("PUBLISH ok — command executor authenticated")
        _            <- ZIO.logInfo("Attempting SUBSCRIBE against auth-protected Redis...")
        result       <- subscription.subscribeSingleWith("ch")()
                          .returning[String].runHead
                          .timeout(5.seconds)
                          .tapError(e => ZIO.logError(s"SUBSCRIBE failed: $e"))
        _            <- ZIO.logInfo(s"SUBSCRIBE returned: $result")
      yield ()
    program.provide(layer)
```

## Expected
`SUBSCRIBE` succeeds on the authenticated connection (parity with `PUBLISH`).

## Actual
```
PUBLISH ok — command executor authenticated
Attempting SUBSCRIBE against auth-protected Redis...
zio.redis.internal.SingleNodeRunner — Reconnecting due to error: ProtocolError: Authentication required.
SUBSCRIBE failed: ProtocolError: Authentication required.
```
The `SingleNodeRunner` reconnect loop retries forever if the app keeps running.

## Root cause
Compare `SingleNodeExecutor.scala:40-47, 93` (authenticates on create and
after `connection.reconnect`) with `SingleNodeSubscriptionExecutor.scala` —
no `AUTH` call anywhere in the file.

## Fix sketch
Mirror `SingleNodeExecutor` in `SingleNodeSubscriptionExecutor`:
thread `RedisConfig` into the executor, issue `AUTH` on the subscription
connection before the `run` fiber starts. Reconnect-time re-auth is a
trickier follow-up because the subscription `send` fiber races on the socket
during `onError`.

I have a patch + tests ready and will open a PR referencing this issue if
you'd like me to take it.

---
_Disclosure: The diagnosis, reproducer, patch, and tests were produced with
heavy assistance from [Claude Code](https://claude.com/claude-code). I've
reviewed all of it and run it end-to-end, but wanted to flag the assist._

loathwine:
## Summary
`SingleNodeSubscriptionExecutor` never issued `AUTH` on the subscription
connection. #1205 fixed this for the command executor but not for
subscriptions. Connecting `RedisSubscription.singleNode` to a Redis with
`requirepass`/ACLs fails every subscribe with `NOAUTH Authentication required.`
and the reconnect loop in `SingleNodeRunner` spams warnings forever.

Mirror `SingleNodeExecutor.auth` on the subscription executor: when
`RedisConfig.auth` is set, synchronously write `AUTH` and await the reply
*before* the `run` fiber starts, so the send/receive loop never races on
the reply.

## Out of scope
Reconnect-time re-auth. The existing subscription `onError` already doesn't
call `connection.reconnect`, and adding both here would race with the
concurrent `send` fiber writing to the same socket. Left for a follow-up
once writes can be serialized (e.g. a connection-write mutex, or adding an
AUTH variant to the subscription `Request` ADT).

## Test plan
- [x] New integration tests in `ConnectionSpec`:
  - `subscription auth required error` — asserts NOAUTH propagates when no auth configured
  - `subscription automatic auth` — subscribe + publish roundtrip with AUTH configured
- [x] `sbt fmt`, `sbt lint`, `sbt client/test` (299 unit tests) green locally
- [x] Manually verified end-to-end against `redis:7 --requirepass ...` using the repro from #1249

Fixes #1249.

---
_Disclosure: The diagnosis, reproducer, patch, and tests were produced with
heavy assistance from [Claude Code](https://claude.com/claude-code). I've
reviewed all of it and run it end-to-end, but wanted to flag the assist._

### DIFF
--- a/modules/redis/src/main/scala/zio/redis/internal/SingleNodeSubscriptionExecutor.scala
+++ b/modules/redis/src/main/scala/zio/redis/internal/SingleNodeSubscriptionExecutor.scala
@@ -203,11 +203,11 @@ private[redis] final class SingleNodeSubscriptionExecutor private (
    * the `run` fiber starts, so that no other effect is reading from or writing to the
    * connection concurrently.
    */
-  val auth: UIO[Unit] = 
+  val auth: UIO[Unit] =
     ZIO.foreachDiscard(config.auth) { creds =>
       val cmd   = RedisCommand("AUTH", AuthInput, UnitOutput).resp(zio.redis.Auth(creds.username, creds.password))
       val bytes = RespValue.Array(cmd.args.map(_.value)).asBytes
-      
+
       val effect =
         for {
           _        <- connection.write(bytes).mapError(RedisError.IOError.apply)
@@ -224,7 +224,7 @@ private[redis] final class SingleNodeSubscriptionExecutor private (
                           ZIO.fail(RedisError.ProtocolError(s"Unexpected response to AUTH: $other"))
                       }
         } yield ()
-      
+
       effect.catchAll(e => ZIO.logError(s"Failed to authenticate subscription connection: $e"))
     }
 }
--- a/modules/redis/src/main/scala/zio/redis/internal/SingleNodeSubscriptionExecutor.scala
+++ b/modules/redis/src/main/scala/zio/redis/internal/SingleNodeSubscriptionExecutor.scala
@@ -203,7 +203,7 @@ private[redis] final class SingleNodeSubscriptionExecutor private (
    * the `run` fiber starts, so that no other effect is reading from or writing to the
    * connection concurrently.
    */
-  val auth: UIO[Unit] = ZIO.foreachDiscard(config.auth) { creds =>
+  val auth = ZIO.foreachDiscard(config.auth) { creds =>
     val cmd                          = RedisCommand("AUTH", AuthInput, UnitOutput).resp(zio.redis.Auth(creds.username, creds.password))
     val bytes                        = RespValue.Array(cmd.args.map(_.value)).asBytes
     val effect: IO[RedisError, Unit] =
--- a/modules/redis/src/main/scala/zio/redis/internal/SingleNodeSubscriptionExecutor.scala
+++ b/modules/redis/src/main/scala/zio/redis/internal/SingleNodeSubscriptionExecutor.scala
@@ -18,16 +18,17 @@ package zio.redis.internal
 
 import zio._
 import zio.concurrent.ConcurrentMap
-import zio.redis.Input.{CommandNameInput, StringInput}
-import zio.redis.Output.PushMessageOutput
+import zio.redis.Input.{AuthInput, CommandNameInput, StringInput}
+import zio.redis.Output.{PushMessageOutput, UnitOutput}
 import zio.redis.api.Subscription
 import zio.redis.internal.PubSub.{PushMessage, SubscriptionKey}
 import zio.redis.internal.SingleNodeRunner.True
 import zio.redis.internal.SingleNodeSubscriptionExecutor.Request
-import zio.redis.{Input, RedisError}
+import zio.redis.{Input, RedisConfig, RedisError}
 import zio.stream._
 
 private[redis] final class SingleNodeSubscriptionExecutor private (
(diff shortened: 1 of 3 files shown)


loathwine:
FYI the failing `publishLocal` / `ci` checks are unrelated to this PR — it's the Docusaurus website build failing with a `ProgressPlugin` webpack-schema mismatch. All 9 test matrix variants (JDK 17/21/25 × Scala 2.12/2.13/3.x) and `lint` pass.

Same failure pattern on unrelated PRs going back to 2026-04-13, e.g. #1248 (scala-steward, no source changes).

Happy to rebase on any docs fix if one lands before review.

modules/redis/src/main/scala/zio/redis/internal/SingleNodeSubscriptionExecutor.scala
```diff
@@ -195,6 +196,34 @@ private[redis] final class SingleNodeSubscriptionExecutor private (
...
     } yield ()
   }
+
+  /**
+   * Authenticates the subscription connection synchronously by writing an `AUTH`
+   * command and awaiting the reply on the same connection. It must be invoked before
+   * the `run` fiber starts, so that no other effect is reading from or writing to the
+   * connection concurrently.
+   */
+  val auth: UIO[Unit] = ZIO.foreachDiscard(config.auth) { creds =>
```
mijicd:
```suggestion
  val auth = ZIO.foreachDiscard(config.auth) { creds =>
```

modules/redis/src/main/scala/zio/redis/internal/SingleNodeSubscriptionExecutor.scala
```diff
@@ -195,6 +196,34 @@ private[redis] final class SingleNodeSubscriptionExecutor private (
...
+    val cmd                          = RedisCommand("AUTH", AuthInput, UnitOutput).resp(zio.redis.Auth(creds.username, creds.password))
+    val bytes                        = RespValue.Array(cmd.args.map(_.value)).asBytes
+    val effect: IO[RedisError, Unit] =
```
mijicd:
```suggestion
    val effect =
```

mijicd (dismissed):

mijicd (approved):

